PR #9, opened 2014-12-17
Update English and French Localization Files

jasonleenaylor:


DistFiles/localizations/Chorus.en.tmx
```diff
@@ -1,922 +1,945 @@
...
+        <seg>&amp;Add &amp;&amp; &amp;OK</seg>
+      </tuv>
+    </tu>
+    <tu tuid="AnnotationEditorView.AddAndOK">
+      <prop type="x-nolongerused">true</prop>
+      <tuv xml:lang="en">
+        <seg>&amp;Add &amp;&amp; &amp;OK</seg>
+      </tuv>
+    </tu>
+    <tu tuid="AnnotationEditorView.AddAndOk">
```
jasonleenaylor: I removed this line and it was re-added as is by the Update code. There must be a button in the actual dialog code named AddAndOK with Cancel as the text. I'm not going to clean up chorus dialogs as part of this change. Good catch though.

papeh:
Glad we're now using pure spaces for XML indents.  Changes look good. :+1:

PR #10, opened 2015-01-29
Update to FW 9 data model of 7000069 & update dependency file

regnrand:


regnrand:
I'll withdraw this for now. I woke up in the middle of the night thinking about more changes that would be needed. I'll resubmit it later.

PR #11, opened 2015-01-29
Remove DownloadFiles target & update Bash script for Windows dependencie...

regnrand:
...s


papeh:
No download_dependencies_linux.sh?

Even if it's not the best name, I'd prefer this script maintain the buildupdate[<platform>].sh name for consistency with other repositories.  However, if you do change the name of the script, please update the ReadMe.

regnrand:
Not yet on the Linux version of the file. There is no working Team City build configuration set up for Linux for FLEx Bridge master. Nor do I see any build configurations for Chorus or Palaso that don't do that new strong name business for Linux.

I'll look over the read me file.

ermshiperete:
Strong names don't seem to be a problem on Linux, so there doesn't need to be special treatment with/without a strong name.

regnrand:
I fixed the partially setup TC build config for FLEx Bridge master Linux build. I then worked up the Linux download script and updated the readme.

papeh:
Thanks for the work in ReadMe; it looks a lot cleaner now.
:+1:

jasonleenaylor:
Hmm, Eberhard's comment makes me think that Mono doesn't honor strong naming, which might cause some issues with Paratext/FLEx interaction on linux. But that isn't an issue here. I like your name for the scripts better than the original.

PR #12, opened 2015-01-29
Update to FW 9 data model of 7000069

regnrand:


papeh:
Strange that some of the numbers were changed down instead of up.
I don't entirely understand what's going on, but it looks ok to me.

regnrand:
internal const int ExpectedExtensionCount = 25;

This was lowered by 1, because the file handler type that supported the now removed archived Scripture texts was removed.

ImportType values had to be set to the new bounds (0 or 1), where before the model change there were 0->3 types of supported import types for Scripture. The two that remain are "Unknown", and "PT6" usfm supported import data types for Scripture.

PR #13, opened 2015-03-19
Use NuGet for the NUnit dependencies

jasonleenaylor:
* Enable package restore
* Use NuGet for each test project that uses NUnit
* Add NuGet targets and depend on the NuGet Restore
  for all test projects


jasonleenaylor:
This will get the master builds of FLExBridge working again

ermshiperete:
The msbuild integrated approach is no longer recommended by the NuGet developers. The other approach doesn't require a change in every .csproj file. 

Possibly you had a `.nuget` subdirectory when you opened the solution - this causes VS to use the msbuild integrated approach and modify the .csproj files.

.nuget/NuGet.targets
```diff
@@ -0,0 +1,93 @@
+<?xml version="1.0" encoding="utf-8"?>
```
ermshiperete: In order to NOT use the msbuild integrated approach, this file has to reside somewhere else, e.g. as `build/NuGet.targets`

FLExBridge VS2010.sln
```diff
@@ -30,6 +30,13 @@ Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "TriboroughBridge-ChorusPlug
 EndProject
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "RepositoryUtility", "src\RepositoryUtility\RepositoryUtility.csproj", "{FC668C72-7472-4774-8501-8E021ECF3BEA}"
 EndProject
+Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = ".nuget", ".nuget", "{E2661AE6-A7BA-41BA-A2CF-DECB376B2729}"
```
ermshiperete: Not necessary for Automatic Package Restore approach

src/FLEx-ChorusPluginTests/FLEx-ChorusPluginTests.csproj
```diff
@@ -1,263 +1,275 @@
-<?xml version="1.0" encoding="utf-8"?>
```
ermshiperete: Changes not necessary for Automatic Package Restore approach

jasonleenaylor: I don't think this is true. For it to work from Visual Studio the package.config file needs to be in the solution somewhere.

jasonleenaylor: Also, the reference must change to the new location for the nunit dependency.

PR #14, opened 2015-03-27
Merge beta branch into master

jasonleenaylor:
There were some commits on Beta that had not yet made there way into master. This will resolve that.


jasonleenaylor:
Reviewers need only to focus on the Merge 'Beta' into master commit unless they see something that they would object to coming over from Beta.

regnrand:
Looks like the merge is fine. I was mostly interested in the installer issues and related readme files. Theory has it that the FB master installer will be put up in a folder (other than "Beta") on the FB download page.

PR #16, opened 2015-03-31
Add a bit more polish to the Installer

jasonleenaylor:
* Simplifly the license rtf to avoid a WIX bug where it failed
  to display
* Add custom banner and background with SIL branding
* Change company info to SIL International


regnrand:
LGTM.

PR #19, opened 2015-05-01
Update debian dependencies to build with mono-sil and use curl

jasonleenaylor:
* update debian dependencies with mono-sil and geckofx29
* update run-app script to use new mono and gecko
* modify the make file to use the download dependencies script
* install MercurialExtensions

### DIFF
--- a/lib/common/run-app
+++ b/lib/common/run-app
@@ -16,18 +16,20 @@ ARCH=$(uname -m)
 # Dependency locations
 # search for xulrunner and geckofx, select the best, and add its location to LD_LIBRARY_PATH
 # (the search order must be the same as in Geckfx-Core/XULRunnerLocator.cs)
-XULRUNNER=/usr/lib/xulrunner-geckofx
-if [ ! -d ${XULRUNNER} ]; then XULRUNNER=/usr/lib/xulrunner-14.0.1; fi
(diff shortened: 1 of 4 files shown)


lib/common/run-app
```diff
@@ -36,9 +38,12 @@ else
...
 
 # Add packaged mono items to paths
-PATH="${MONO_PREFIX}/bin:${PATH}"
-LD_LIBRARY_PATH="${MONO_PREFIX}/lib:${LD_LIBRARY_PATH}"
+
 PKG_CONFIG_PATH="${MONO_PREFIX}/lib/pkgconfig:${PKG_CONFIG_PATH}"
+PATH="/opt/mono-sil/bin:${PATH}"
+LD_LIBRARY_PATH="/opt/mono-sil/lib:${LD_LIBRARY_PATH}"
+PKG_CONFIG_PATH="/opt/mono-sil/lib/pkgconfig:${PKG_CONFIG_PATH}"
+MONO_GAC_PREFIX="/opt/mono-sil:/usr"
```
mccarthyrb: The above /opt/mono-sil paths would ignore the /usr/local path if that is the mono being used. It would be better to use the variable ${MONO_PREFIX} for PATH, LD_LIBRARY_PATH, PKG_CONFIG_PATH, and MONO_GAC_PREFIX.

PR #20, opened 2015-05-04
Update flexbridge Beta to work with tagged ICU build

jasonleenaylor:
* Use the pinned and tagged icu40 build from TC
* Change suggests to require for fieldworks and restrict
  the maximum version.


regnrand:
LGTM.

PR #24, opened 2015-06-25
Useful Message when run directly (LT-16214)

papeh:
When FLEx Bridge is run w/o command line args, assume
the user double-clicked and show a useful message.
Do not green-screen in this case.


src/TriboroughBridge-ChorusPlugin/CommandLineProcessor.cs
```diff
@@ -84,7 +84,7 @@ public static class CommandLineProcessor
 							//This is an unparsable command line
 							Console.WriteLine(CommonResources.kInvalidCommandLineOptions);
 							//Signal FLEx or other apps
-							throw new ApplicationException(String.Format("Invalid command line options: {0}.", args));
+							throw new ApplicationException(String.Format("Invalid command line options: {0}. Please launch from FLEx.", args));
```
papeh: Unless there are objections, I'll leave this message intact, since it handles a different case.

Incidentally, this block could use more work (Console.WriteLine doesn't get a chance to do anything before we throw, and `args` shows up as 'String[]' in the exception message), but I'm not sure if it would be worth the effort to fix that.

regnrand: Strictly speaking, FB was designed to be run by any number of clients, with FLEx being one of them. So, it would be better to not tell users to run FB from FLEx. (Yes, the design may not have been implemented in all cases with multiple clients in mind, but that was the design.)

jasonleenaylor:
LGTM

PR #29, opened 2015-08-07
Add pkg-config and curl Build-Depends

marksvc:
* Since pkg-config is needed, and wget has trouble with a certificate.

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -7,6 +7,8 @@ Build-Depends: debhelper (>= 8.0.0),
  libmono-winforms2.0-cil,
  libmono-cairo2.0-cil,
  geckofx29, unzip, wget,
+ pkg-config,
+ curl,
  adduser (>= 3.11)
 Standards-Version: 3.9.2
 Homepage: http://projects.palaso.org/projects/fwbridge


debian/control
```diff
@@ -7,6 +7,8 @@ Build-Depends: debhelper (>= 8.0.0),
  libmono-winforms2.0-cil,
  libmono-cairo2.0-cil,
  geckofx29, unzip, wget,
+ pkg-config,
+ curl,
```
jasonleenaylor: why does it need wget and curl?

neilmayhew:
:+1:

debian/control
marksvc: It downloads various dependencies.

marksvc: Oh, it might not need wget AND curl. It didn't occur to me to remove 'wget', but I could if that's thought to be a good idea.

marksvc: I am trying a package build without the 'wget' Build-Depends.

marksvc: The package build fails without 'wget', so let's include both curl and wget.

PR #31, opened 2015-08-12
Revert the Schedule attribute change and add dev target

jasonleenaylor:
The Chorus merge module will not function correctly if the Schedule
attribute is anything but afterInstallValidate

* Reset 'Schedule' to afterInstallValidate
* Add DevelopInstaller target and comment for how to validate installer
  work on a developer machine


papeh:
:+1:

PR #36, opened 2015-08-18
Add unzip to binary dependency

marksvc:
posinst requires unzip to unpack mercurial.

(cherry picked from commit fcca735fc0950a601e92fbd07fcefe31ac65a9d1)

Conflicts:

	debian/control


StephenMcConnel:
LGTM. :+1:

jasonleenaylor:
Just confirming you want this in the Beta or 2.2.15(6?) package.

marksvc:
I do want it in the Beta branch/package. It is not super important but sometimes I run into an error associated with its absence. I could live with not incorporating the fix, but it may be helpful now and then if we patch it.

PR #37, opened 2015-08-19
Add the fixutf8 plugin to Mercurial.ini in postinst

jasonleenaylor:
The plugin is required on linux after the upgrade to mercurial 3
in order to handle integer only branch names.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -25,6 +25,7 @@ case "$1" in
 			ARCH="$(/usr/bin/arch)"
 			unzip Mercurial-${ARCH}.zip
 			chmod +x Mercurial/hg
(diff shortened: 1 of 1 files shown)


marksvc:
LGTM. However: 
* Don't forget spaces around the '=' in the extension setting if that's needed. Maybe not.
* The test for unzipping Mercurial makes me wonder if it will not re-unzip mercurial on upgrades.

jasonleenaylor:
the whitespace is not important, I was wondering about the unzip test also.

jasonleenaylor:
I've decided to leave the mercurial unzip on upgrade enhancement until later.

PR #38, opened 2015-08-19
Add GeckoBrowserAdapter which is needed at runtime on Linux

jasonleenaylor:


marksvc:
+1. Though I hope your editing of the generated file was using the generating tool, rather than manual.

jasonleenaylor:
@marksvc Yes, it was generated running the buildupdate.rb from the changes I had already made on TeamCity.

PR #41, opened 2015-08-24
Merge in beta

jasonleenaylor:


src/Installer/ReleaseNotes.md
```diff
@@ -1,6 +1,8 @@
 ## DEV_VERSION_NUMBER: DEV_RELEASE_DATE
 * Chorus: Upgrade Mercurial to version 3
-## 2.2.14 23/Jan/2015
+## 2.2.15 4/May/2015
+* Chorus: Fix overflow error when receiving very large project data with resumable option
+## 2.2.14 3/Feb/2015
```
regnrand: Is the Jan or Feb date right for 2.2.14? Seems odd to reset it at this late date.

src/Installer/Beta.htm
```diff
@@ -9,6 +9,9 @@
 		<h4>Release Notes</h4>
 		<ul>
 			<li>DEV_VERSION_NUMBER<ol>
+				<li>Chorus: Fix overflow error when receiving very large project data with resumable option.</li>
```
regnrand: Seems like you need to move "Fix overflow" into 2.2.15 and then add "Chorus: Upgrade Mercurial to version 3" as the current version.

src/Installer/ReleaseNotes.md
jasonleenaylor: It got reset at a correct right time in the Beta branch, it just needed to be merged into master.

src/Installer/Beta.htm
jasonleenaylor: Well, this is for the Beta, so I think it is technically correct. At some point we need to declare that Beta was actually stable. The master branch is currently claiming to be Alpha, but should soon be Beta. We should do it in some sort of graduation ceremony, maybe play "Pomp and Circumstance"

marksvc:
I am doing a quick look at the changes to Makefile and debian/, and I don't see anything that stands out as a problem. I did notice the specification of the fieldworks-applications dependency as '>> 8.1.9'. Probably '>> 8.2.0~' or '>> 8.2' would be better but it won't make a difference if we never release a FW 8.1.10. 

I noticed the removal of the mono-devel Build-Depend but that sounds familiar.

I am trying the package build on my own system to see if it has any trouble.

marksvc:
I was able to build a package for Ubuntu 14.04 trusty 64-bit, though I found I needed to include ca-certificates as a Build-Depend. Not sure if that's specific to my machine. You could go with the code as-is, and we could add ca-certificates as a Build-Depend after the fact.

PR #43, opened 2015-09-03
Update changelog for 2.3.5 release package

jasonleenaylor:


marksvc:
I think you should change 'UNRELEASED' to 'unstable' and priority from 'medium' to 'low'. Then LGTM.

PR #44, opened 2015-09-04
Install ChorusHub on linux until a separate package is created

jasonleenaylor:
* The shortcut had not yet been removed and we don't have a package
  for ChorusHub yet. So until we can make one include ChorusHub.exe
  so that the shortcut will launch the service.


papeh:
LGTM :shipit:

marksvc:
Sounds good.

PR #48, opened 2015-09-12
Add ca-certificates build-depend

marksvc:
* Needed by at least one machine's pbuilder.


marksvc:
This doesn't seem necessary on the server, but whenever I try packaging flexbridge on my own machine I need ca-certificates as a build-depends.

PR #50, opened 2015-09-17
Make chorushub run as a service

mccarthyrb:
The launcher needs to use "mono-service" to launch the
program as a daemon on linux. (LT-16587)


jasonleenaylor:
In a future change it would be good to refactor the common library setting code into a shared shell script for run-app and run-app-service to use. This fixes the problem and I'd like to see it tested before any further enhancement.

PR #51, opened 2015-09-17
Update for chorushub service to use the same script

mccarthyrb:
Refactored script to allow for fieldworks-chorushub to pass a
"--service" argument that run-app processes and changes the
command from mono to mono-service.


marksvc:
+1 as long as you have tested the script with --service not being specified.

mccarthyrb:
Need to update the run-app script a bit more. There needs to be a shift to remove the "--service" argument before executing the command.

jasonleenaylor:
LGTM

PR #54, opened 2015-09-25
Update RepositoryUtility to be able to downgrade hgrc files

jasonleenaylor:
* Add menu item for projects with as sub item for the Downgrade
* Add the downgrade hgrc function


src/RepositoryUtility/RevertHgrcProjectFolderForm.cs
```diff
@@ -0,0 +1,81 @@
...
+			PopulateList(projectsLocation);
+			_okButton.Enabled = _listView.Items.Count > 0;
+		}
+
+		private void PopulateList(string projectsLocation)
+		{
+			_listView.SuspendLayout();
+			_listView.Items.Clear();
+			if(!Directory.Exists(projectsLocation))
+				return;
```
papeh: If we return here, the layout is still suspended.

PR #55, opened 2015-09-25
Update release documentation, increase version to 2.3.9

jasonleenaylor:
* Update Alpha.htm and ReleaseNotes.md with some historical information
* Update the ReadMe.md with release instructions
* Update the Makefile to generate the right version number
* Update the build.common.proj even though it isn't necessary
* Update the appcast.xml so that the next Alpha release should be
  offered to users automatically.
* Update the index.html to offer information about both Beta and Alpha
  releases

### DIFF
--- a/src/Installer/Alpha.htm
+++ b/src/Installer/Alpha.htm
@@ -8,6 +8,16 @@
 		<h4>Release Notes (Preview of things to come)</h4>
 		<ul>
 			<li>DEV_VERSION_NUMBER<ol>
+				<li>Chorus: Update documentation of Chorus Hub</li>
+				<li>FLExBridge: Update documentation of recent fixes</li>
+			</ol></li>
+			<li>2.3.8<ol>
+				<li>Chorus: Make large internet Send/Receive options more robust (Warning instead of error when retrying a failed Chunk)</li>
+				<li>Chorus: Improve Chorus Hub Instructions</li>
+				<li>FLExBridge: Provide Chorus Hub Service option on Linux</li>
+				<li>FLExBridge: Various linux installer enhancements</li>
+			</ol></li>
+			<li>2.3.5<ol>
 				<li>L10NSharp: Improve debug information for linux localization</li>
 			</ol></li>
 			<li>2.3.2<ol>
--- a/src/Installer/ReleaseNotes.md
+++ b/src/Installer/ReleaseNotes.md
@@ -1,4 +1,12 @@
 ## DEV_VERSION_NUMBER: DEV_RELEASE_DATE
+* Chorus: Update documentation of Chorus Hub
+* FLExBridge: Update documentation of recent fixes
+## 2.3.8 24/Sep/2015
+* Chorus: Make large internet Send/Receive options more robust (Warning instead of error when retrying a failed Chunk)
+* Chorus: Improve Chorus Hub Instructions
+* FLExBridge: Provide Chorus Hub Service option on Linux
+* FLExBridge: Various linux installer enhancements
+## 2.3.5 3/Sep/2015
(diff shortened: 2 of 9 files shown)


src/Installer/Alpha.htm
```diff
@@ -8,6 +8,16 @@
 		<h4>Release Notes (Preview of things to come)</h4>
 		<ul>
 			<li>DEV_VERSION_NUMBER<ol>
+				<li>Chorus: Update documentation of Chorus Hub</li>
+				<li>FLExBridge: Improve developer documentation and update documentation of recent fixes</li>
```
papeh: this isn't user-facing

src/Installer/ReleaseNotes.md
```diff
@@ -1,4 +1,12 @@
 ## DEV_VERSION_NUMBER: DEV_RELEASE_DATE
+* Chorus: Update documentation of Chorus Hub
+* FLExBridge: Improve developer documentation and update documentation of recent fixes
```
papeh: this improvement isn't user-facing (but I'm not sure that this file is, either, is it?

src/Installer/appcast.xml
```diff
@@ -8,9 +8,9 @@
 		 <item>
 			<title>Version DEV_VERSION_NUMBER</title>
 						<sparkle:releaseNotesLink>
-							http://downloads.palaso.org/FlexBridge/Beta/ReleaseNotes.md
+							http://downloads.palaso.org/FlexBridge/Alpha/ReleaseNotes.md
 						</sparkle:releaseNotesLink>
-			<enclosure url="http://downloads.palaso.org/FlexBridge/Beta/FLExBridgeInstaller.msi" sparkle:version="DEV_VERSION_NUMBER" type="application/octet-stream" />
+			<enclosure url="http://downloads.palaso.org/FlexBridge/Alpha/FLExBridgeInstaller.msi" sparkle:version="DEV_VERSION_NUMBER" type="application/octet-stream" />
```
papeh: don't forget to say something about this when you know more.

src/Installer/ReleaseNotes.md
regnrand: Yes, users will see this file. Sparkle will show it to users before they install it, so they can decide if they want what the new version contains.

src/Installer/Alpha.htm
regnrand: Users can see it on the FB download web page:

http://downloads.palaso.org/FlexBridge/index.htm

src/Installer/appcast.xml
regnrand: I would expect to see two channels: Alpha & Beta.

src/Installer/index.htm
```diff
@@ -26,6 +26,11 @@
 			<tbody>
 				<tr>
 					<td id="beta" >
+						<iframe id="Beta" src="Beta/Beta.htm" title="FLEx Bridge">
+							<p>Your browser does not support iframes.</p>
+						</iframe>
+					</td>
+					<td id="alpha" >
```
regnrand: Shouldn't the Alpha column come first on the website?

jasonleenaylor: I was thinking that they would be listed from more to less stable in left to right order. So, I'm open to suggestions, but I did do it on purpose.

regnrand: That's fine.

src/Installer/appcast.xml
jasonleenaylor: So only one channel per appcast is supported by NetSparkle. I've made a change to copy this into the UploadFolder so that we can have different appcast.xml for the different release types from this point forward.

papeh:
:+1:

PR #60, opened 2015-10-20
Move the stamped about into the output folder during the build

jasonleenaylor:
[Review on Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/60)

### DIFF
--- a/build/FLExBridge.build.mono.proj
+++ b/build/FLExBridge.build.mono.proj
@@ -20,7 +20,7 @@
 
   <Import Project="$(RootDir)/build/build.common.proj"/>
 
-  <Target Name="Build" DependsOnTargets="Clean;CreateDirectories;SetAssemblyVersion;UnzipMercurial;Compile">
+  <Target Name="Build" DependsOnTargets="Clean;CreateDirectories;SetAssemblyVersion;UnzipMercurial;Compile;CopyAbout">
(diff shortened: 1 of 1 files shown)


marksvc:
Looks okay. Though I would have expected this to be part of something more generic. I wonder if some of the recent code removed, about stamping, had been doing this for us previously.

Also, will Installer\ReleaseNotes.md need to be part of this? I guess we probably don't ship that. Maybe a ReleaseNotes.html file?


---

Reviewed 1 of 1 files at r1.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/60)

marksvc:
From my limited familiarity with these files I will say LGTM,  unless we need to also include ReleaseNotes.


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/60)

jasonleenaylor:
We weren't shipping anything else on Linux yet, but we might want to look at that in the future. The stamping of the about.htm was being done twice before, I removed the linux side of it, but the common stamping was only applying to the Installer location. This copy function is an alternative to stamping the file in different places for the different operating systems.


---

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/60)

PR #67, opened 2015-12-07
LT-16773: Include new Dictionary Configs in Send/Receive

papeh:
TODO: special merge step for *.fwdictconfig

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/67)

### DIFF
--- a/src/FLEx-ChorusPlugin/FLEx-ChorusPlugin.csproj
+++ b/src/FLEx-ChorusPlugin/FLEx-ChorusPlugin.csproj
@@ -1,158 +1,158 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
 <Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-	<Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
-	<Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
-	<ProductVersion>8.0.30703</ProductVersion>
-	<SchemaVersion>2.0</SchemaVersion>
-	<ProjectGuid>{133981BF-62BF-453A-8ED0-AEFA386FEBB8}</ProjectGuid>
-	<OutputType>Library</OutputType>
-	<AppDesignerFolder>Properties</AppDesignerFolder>
-	<RootNamespace>FLEx_ChorusPlugin</RootNamespace>
-	<AssemblyName>FLEx-ChorusPlugin</AssemblyName>
-	<TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
-	<FileAlignment>512</FileAlignment>
+    <Configuration Condition=" '$(Configuration)' == '' ">Debug</Configuration>
+    <Platform Condition=" '$(Platform)' == '' ">AnyCPU</Platform>
+    <ProductVersion>8.0.30703</ProductVersion>
+    <SchemaVersion>2.0</SchemaVersion>
+    <ProjectGuid>{133981BF-62BF-453A-8ED0-AEFA386FEBB8}</ProjectGuid>
+    <OutputType>Library</OutputType>
+    <AppDesignerFolder>Properties</AppDesignerFolder>
+    <RootNamespace>FLEx_ChorusPlugin</RootNamespace>
+    <AssemblyName>FLEx-ChorusPlugin</AssemblyName>
+    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
+    <FileAlignment>512</FileAlignment>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Debug|AnyCPU' ">
-	<DebugSymbols>true</DebugSymbols>
-	<DebugType>full</DebugType>
-	<Optimize>false</Optimize>
-	<OutputPath>..\..\output\Debug\</OutputPath>
-	<DefineConstants>DEBUG;TRACE</DefineConstants>
-	<ErrorReport>prompt</ErrorReport>
-	<WarningLevel>4</WarningLevel>
-	<PlatformTarget>x86</PlatformTarget>
-	<TargetFrameworkProfile>Client</TargetFrameworkProfile>
-	<UseVSHostingProcess>false</UseVSHostingProcess>
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>..\..\output\Debug\</OutputPath>
+    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <PlatformTarget>x86</PlatformTarget>
+    <TargetFrameworkProfile>Client</TargetFrameworkProfile>
+    <UseVSHostingProcess>false</UseVSHostingProcess>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(Configuration)|$(Platform)' == 'Release|AnyCPU' ">
-	<DebugType>pdbonly</DebugType>
-	<Optimize>true</Optimize>
-	<OutputPath>..\..\output\Release\</OutputPath>
-	<DefineConstants>TRACE</DefineConstants>
-	<ErrorReport>prompt</ErrorReport>
-	<WarningLevel>4</WarningLevel>
-	<PlatformTarget>x86</PlatformTarget>
-	<TargetFrameworkProfile>Client</TargetFrameworkProfile>
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>..\..\output\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <PlatformTarget>x86</PlatformTarget>
+    <TargetFrameworkProfile>Client</TargetFrameworkProfile>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'ReleaseMono|AnyCPU'">
-	<OutputPath>..\..\output\ReleaseMono\</OutputPath>
-	<DefineConstants>TRACE;MONO</DefineConstants>
-	<Optimize>true</Optimize>
-	<DebugType>pdbonly</DebugType>
-	<ErrorReport>prompt</ErrorReport>
-	<WarningLevel>4</WarningLevel>
+    <OutputPath>..\..\output\ReleaseMono\</OutputPath>
+    <DefineConstants>TRACE;MONO</DefineConstants>
+    <Optimize>true</Optimize>
+    <DebugType>pdbonly</DebugType>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'DebugMono|AnyCPU'">
-	<DebugSymbols>true</DebugSymbols>
-	<OutputPath>..\..\output\DebugMono\</OutputPath>
-	<DefineConstants>TRACE;DEBUG;MONO</DefineConstants>
-	<DebugType>full</DebugType>
-	<UseVSHostingProcess>false</UseVSHostingProcess>
-	<ErrorReport>prompt</ErrorReport>
-	<WarningLevel>4</WarningLevel>
+    <DebugSymbols>true</DebugSymbols>
+    <OutputPath>..\..\output\DebugMono\</OutputPath>
+    <DefineConstants>TRACE;DEBUG;MONO</DefineConstants>
+    <DebugType>full</DebugType>
+    <UseVSHostingProcess>false</UseVSHostingProcess>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
   </PropertyGroup>
   <ItemGroup Condition="'$(Configuration)' == 'Release'">
-	<Reference Include="Chorus">
-	  <HintPath>..\..\lib\Release\Chorus.exe</HintPath>
-	</Reference>
-	<Reference Include="ChorusMerge">
-	  <SpecificVersion>False</SpecificVersion>
-	  <HintPath>..\..\lib\Release\ChorusMerge.exe</HintPath>
-	</Reference>
-	<Reference Include="LibChorus">
-	  <SpecificVersion>False</SpecificVersion>
-	  <HintPath>..\..\lib\Release\LibChorus.dll</HintPath>
-	</Reference>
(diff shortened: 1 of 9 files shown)


regnrand:
Reviewed 5 of 5 files at r1.

---

*[src/FLEx-ChorusPlugin/Infrastructure/FlexFolderSystem.cs, line 65 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K4ykN4DDIWQ5V9BbyKk:-K4ykN4DDIWQ5V9BbyKl:-1667123787) ([raw file](https://github.com/sillsdev/flexbridge/blob/6cdce61cea4502ac2e0fbb828a7dbf9c08bc014c/src/FLEx-ChorusPlugin/Infrastructure/FlexFolderSystem.cs#L65)):*
Are the config files deeper than the parent folder (ConfigurationSettings)?  If not, then you only need "*.fwdictconfig".

With no definitions set up for the fwdictconfig file(s), you won't be happy with the merge. Why was that left for another day?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/67)*

papeh:

---

*[src/FLEx-ChorusPlugin/Infrastructure/FlexFolderSystem.cs, line 65 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K4ykN4DDIWQ5V9BbyKk:-K52RHoZNAHTGYYa30Py:-2124216168) ([raw file](https://github.com/sillsdev/flexbridge/blob/6cdce61cea4502ac2e0fbb828a7dbf9c08bc014c/src/FLEx-ChorusPlugin/Infrastructure/FlexFolderSystem.cs#L65)):*
yes, at `ConfigurationSettings/(Dictionary|ReversalIndex)/*.fwdictconfig`

After 5pm :smile: -- I'll try to get an amendment in tomorrow.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/67)*

regnrand:

---

*[src/FLEx-ChorusPlugin/Infrastructure/FlexFolderSystem.cs, line 65 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K4ykN4DDIWQ5V9BbyKk:-K52ToRmorLoolaY_ggK:-1782123643) ([raw file](https://github.com/sillsdev/flexbridge/blob/6cdce61cea4502ac2e0fbb828a7dbf9c08bc014c/src/FLEx-ChorusPlugin/Infrastructure/FlexFolderSystem.cs#L65)):*
Ok, on using "**" for the nested files. Waiting for the  smarts for the merge....

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/67)*

regnrand:
Reviewed 3 of 5 files at r2.
Review status: 5 of 7 files reviewed at latest revision, 6 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 40 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aMua41nDMfYj9dm0v:-K5aMua41nDMfYj9dm0w:688589574) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L40)):*
What other options might there be to not use the public static property, or adding a dependency to some Handler? It will take more thinking, since, almost nobody in FB cares about where Flex is located. FB used to look at the Registry for this, but that has gone away in favor of the command line information.

One bold option would be to put the schema in the project folder, rather than in the app folder.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 20 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aBNd8IQ9v8hdmm3u4:-K5aBNd8IQ9v8hdmm3u5:-1375744919) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L20)):*
Data members tend to start with "_" in FB. That makes them look different than local variables.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 21 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aGGwzbvhHxKB02duI:-K5aGGwzbvhHxKB02duJ:146665147) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L21)):*
You should probably get rid of this property. It isn't adequate to use "../../DistFiles" on developer machines, since we can be working on Release builds, not Debug builds.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 34 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aGtybKICB23MCOq_A:-K5aGtybKICB23MCOq_B:-1169912227) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L34)):*
No need for the safety net. It should just throw if "dictionaryConfigXsdPath" is null, when it checks to see if the file exists. That should keep you honest in making sure the code knows where to find it.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 48 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aR48oy-mHm6LTiQXQ:-K5aR48oy-mHm6LTiQXR:1959439745) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L48)):*
What is the thinking on handling updates to the schema regarding S/R? That is, how will S/R cope with such changes in the schema?

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/FwLinkArgs.cs, line 656 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aDNtU1jaUbSBHag4D:-K5aDNtU1jaUbSBHag4E:-1941800554) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/Handling/FwLinkArgs.cs#L656)):*
Good. As you might have guessed, this code was copied from the FLEx version.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/67)*

regnrand:
Reviewed 2 of 5 files at r2.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/67)*

papeh:

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 40 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aMua41nDMfYj9dm0v:-K5acMAe61rP9A6j7xW0:-669811675) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L40)):*
I knew you wouldn't like this, but I was out of ideas.  I think I like looking in the Registry best of these three options (if that wouldn't add too much extra code).

Mirroring the schema in each project directory would make it slightly more accessible, and we could find it in the validator method from its relative path to the configuration, but I don't like the idea of copying this file around and needing to keep all of its instances up to date.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 48 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aR48oy-mHm6LTiQXQ:-K5aePoyghUP47ADxxNV:22821709) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L48)):*
I'm not sure I understand your question, perhaps because I'm not sure exactly when Chorus validates files.

My understanding is that we would expect all users on a project to be using the same version of the schema, and to upgrade at the same time.  So if the user upgrades FLEx, FLEx would upgrade each project's configurations when it is loaded.  Then, when the project is S/R'd, FB will validate the upgraded configuration against the new schema.  Does that sound about right?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/67)*

regnrand:

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 40 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aMua41nDMfYj9dm0v:-K5aiWW-1KIktxu37d0f:1375190432) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L40)):*
On use of the Registry... We moved a fair bit of heaven and earth to remove using the Registry, when we added all of those command line args to the FB exe. Let's struggle more on how to not re-introduce using it, given how the dumb FW Registry info can be in so many places these days (HKLM and/or HKCU plus version 8 or 9 options for each of those).

I think "TriboroughBridge_ChorusPlugin.Utilities" may work. It already knows too much about the two types of bridges, so one more piece of Flex stuff won't make it any worse, even if it won't make it better. a new static method there can take the arg "commandLineArgs["-f"]" (the FW fixit exe), and work out from there where the schema is. That should be able to find DistFiles on user's machines and on dev machines (both release and debug builds).

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 48 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aR48oy-mHm6LTiQXQ:-K5ansxS5ebyp6jZcnRr:1600903306) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L48)):*
HP: "My understanding...."

We've built S/R (Chorus and FB) to not require users to upgrade at the same time. They won't upgrade at the same time for any number of good reasons, so our code needs to adapt to the given version. (See FB's MetadataCache class in the Flex end to see how FB acts like a chameleon and can adjust itself to work with any version of Flex data.

HP: "So if the user upgrades FLEx, FLEx would upgrade each project's configurations when it is loaded."

Fine. The way FLEx handles this is to add a data migration and bump up its model version number by 1. Then, FB will need its MetadataCache set to match the FLEx model version.  In this case, Flex will only change the schema and config file(s) in a way that suits FLEx. So, the MDC in FB will be a no-op data migration, as will be likely the case in Flex.

Then, a new branch will be created in the Flex proj repo, which will let the users use their own version, as long as they like.

BUT, the smarts in FB that deal with the original schema and any updates that follow and do merges on the version the current user has, which may be an older version than is currently being shipped.

HP: "I'm not sure exactly when Chorus validates files"

Theory has it the validation is done right before a commit. I don't recall if Chorus checks the validation after a merge and before its commit, but a look at the Chorus code will confirm that behavior.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/67)*

papeh:
Review status: 5 of 9 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 40 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5aMua41nDMfYj9dm0v:-K5ffFREOnAKOGM3t-nZ:-1437134065) ([raw file](https://github.com/sillsdev/flexbridge/blob/1ea4d08f20419e5ff93f5b2cb5b516ae79003a28/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L40)):*
Ok, I'll move the static property to TriboroughBridge_ChorusPlugin.Utilities

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/67)*

regnrand:
Reviewed 4 of 4 files at r3.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 22 \[r3\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5fo8QsemmEU6rr9was:-K5fo8QsemmEU6rr9wat:828328640) ([raw file](https://github.com/sillsdev/flexbridge/blob/3a0065d02b9e1166de184049310fe887141e057a/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L22)):*
Any reason this and the data member need to be static now?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/67)*

papeh:

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 22 \[r3\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/67#-K5fo8QsemmEU6rr9was:-K5g91fZOfSvGDAwuEq0:939143033) ([raw file](https://github.com/sillsdev/flexbridge/blob/3a0065d02b9e1166de184049310fe887141e057a/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L22)):*
nope; I'll change it

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/67)*

jasonleenaylor:
Since the strategy isn't being successfully used yet and no crashes result I am happy with this as an intermediate commit. This commit will also verify all the infrastructure that @papeh has set up. A later commit can be made to get this handler used and fill in the unimplemented methods.

PR #69, opened 2015-12-22
LT-16773: S/R New Configs: Implement Merge Step

papeh:
Use file-as-atom merge strategy.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/69)

### DIFF
--- a/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs
+++ b/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs
@@ -1,5 +1,5 @@
 ﻿// --------------------------------------------------------------------------------------------
-// Copyright (C) 2010-2013 SIL International. All rights reserved.
+// Copyright (C) 2010-2016 SIL International. All rights reserved.
 //
 // Distributable under the terms of the MIT License, as specified in the license.rtf file.
 // --------------------------------------------------------------------------------------------
@@ -36,7 +36,7 @@ namespace FLEx_ChorusPlugin.Infrastructure.ActionHandlers
 		{
 			// -p <$fwroot>\foo\foo.fwdata
 			var projectDir = Path.GetDirectoryName(commandLineArgs["-p"]);
-			Utilities.FwAppsDir = commandLineArgs["-fwAppsDir"];
+			EnsureAccessToDictionaryConfigXsd(commandLineArgs);
 			using (var chorusSystem = Utilities.InitializeChorusSystem(projectDir, commandLineArgs["-u"], FlexFolderSystem.ConfigureChorusProjectFolder))
 			{
 				var newlyCreated = false;
@@ -117,6 +117,19 @@ namespace FLEx_ChorusPlugin.Infrastructure.ActionHandlers
 						File.Delete(lockPathname);
 				}
 			}
+		}
+
+		private static void EnsureAccessToDictionaryConfigXsd(IDictionary<string, string> commandLineArgs)
+		{
+			var fwAppsDir = commandLineArgs["-fwAppsDir"];
+			var innerXsdPath = Path.Combine("Language Explorer", "Configuration", SharedConstants.DictConfigSchemaFilename);
+			var xsdPath = Path.Combine(fwAppsDir, innerXsdPath);
+			if (!File.Exists(xsdPath))
+				xsdPath = Path.Combine(fwAppsDir, "..", "..", "DistFiles", innerXsdPath);
+			if (!File.Exists(xsdPath))
+				return;
+			var xsdPathInProject = Path.Combine(Path.GetDirectoryName(commandLineArgs["-p"]), "Temp", SharedConstants.DictConfigSchemaFilename);
+			File.Copy(xsdPath, xsdPathInProject, true);
 		}
 
 		/// <summary>Removes .hg repo and other files and folders created by S/R Project</summary>
--- a/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs
+++ b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs
@@ -1,5 +1,5 @@
 ﻿// --------------------------------------------------------------------------------------------
-// Copyright (C) 2015 SIL International. All rights reserved.
+// Copyright (C) 2015-2016 SIL International. All rights reserved.
 //
 // Distributable under the terms of the MIT License, as specified in the license.rtf file.
 // --------------------------------------------------------------------------------------------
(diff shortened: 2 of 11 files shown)


papeh:
I haven't been able to exercise GetChangePresenter or Find2WayDifferences, but their implementations are trivial.

regnrand:
Reviewed 9 of 9 files at r1.
Review status: all files reviewed at latest revision, 7 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 28 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6BNah62vfMFYOyk17Z:-K6BNah62vfMFYOyk17_:-1351456712) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L28)):*
Are you sure you don't want to copy the xsd into the FW project folder?

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 71 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6BNteQcls_YVWLg0-T:-K6BNteQcls_YVWLg0-U:-762395485) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L71)):*
No. Don't add a Registry call. It took long enough to get rid of them, and FB doesn't want them back in its code. The trouble with the Registry approach is FW keeps changing where it is installed, so the Registry approach has too many options to look at. Such a move would then bind FB to need related changes for every new FW version, and have backwards support for all of them. Technically, FB is backwards compatible in handling data model changes all they way back to FW 7 days (when I started building FB).

As far as devs using "different repo structures".... That is true. I have never used "C:\fwrepo\fw\".

How about having Flex (1st preference) or FB (second choice) move the xsd into the project folder somewhere? FB can be set up to not put it into the repo, and could even delete it, when the S/R operation is done.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 85 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6AyozECwirxIW_ThNI:-K6AyozF0czM9oYMIUfK:1877835299) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L85)):*
While this method did not change in this round of work, it probably should. I think almost every other "CanValidateFile" method in FB starts by calling "FileUtils.CheckValidPathname". This one should call it, as well. It can then also call the check to see if the schema exists.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 104 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6AzsapuPr7wr1apWun:-K6AzsapuPr7wr1apWuo:-1215755266) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L104)):*
The more typical return value in FB at this point is simply: "return e.Message;" I don't object to this message, but then maybe the others should do the same, perhaps in some new, common, method they can all use.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 126 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6B-i1s7NKccSqWM-js:-K6B-i1s7NKccSqWM-jt:1770896016) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L126)):*
The commit comments says: "Use file-as-atom merge strategy." This code doesn't actually do that, but it could. (See below.)

I don't think "UnmergableFileTypeConflict" is the conflict you want. That is only used in Chorus for unknown file extensions. The user will see this message for this conflict: "Chorus did not have the ability to merge both user's version of the file 'PathnameOfFileWithUnknownExtension'." In this case Chorus does indeed have the ability to do the merge, since it is an xml file.

I think there are some boundary/edge cases that need to be supported here.

1) Chorus ethos says that if either user deleted the file, but the other made changes to it then the edit wins, no matter which user it was.

2) The assumption here also seems to be that the changes do in fact conflict, where another option would be both made the same changes. That should be reported as a 'both made the same changes' (I forget the exact class name.) change report, not a conflict report.

3) It could be that both added a new file and there is no common parent. The two additions could be the same or conflicting.

I think you can do the merge using Chorus' merge system and get the right report(s). You can create an instance if the Chorus class XmlMerger, and set it up so that the root element has one and only one ElementStrategy for the document root, which has its "IsAtomic" property set to 'true'.  Once the Xmlmerger instance has been initialized, you can then call its "MergeFiles" method. That method should cover all three of the above edge cases and any other edge cases I may have forgotten to mention. (See Chorus' "MergeAtomicElementService" class to see all possible covered cases. If you can see any cases that are missing, it is a bug in "MergeAtomicElementService". ;-)

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 76 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6BJhgSWuun8sZ5fpSe:-K6BJhgTqQ6Nqggbw0iI:-1341089163) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L76)):*
Technically, the test data isn't well-formed xml, so the test name is mis-leading. Well-formed xml may, or may not, be valid when checked against the XSD/DTD.

---

*[src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs, line 146 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6AuXt9KUjiUYlJZqHz:-K6AuXt9KUjiUYlJZqI-:1400595484) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs#L146)):*
Why was the call to the MDC blocked?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/69)*

papeh:
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 28 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6BNah62vfMFYOyk17Z:-K7CQzwuyX77W_tApe5n:-787028801) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L28)):*
You make it sound so appealing :-)

---

*[src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs, line 146 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6AuXt9KUjiUYlJZqHz:-K6EDLFkNI4NJMm4ALzB:-751335023) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs#L146)):*
`mdc` was unused, and commenting this line doesn't break the test. If you'd like, I'll put it back.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/69)*

papeh:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 85 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6AyozECwirxIW_ThNI:-K7CTQy-FPXBuLbJRFJ6:-499146009) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L85)):*
Done.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 104 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6AzsapuPr7wr1apWun:-K7CWwuLDiIFp751b-l5:1219088880) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L104)):*
Insufficient ROI to change all 28 occurs; returning e.Message (or args.Message).

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 126 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6B-i1s7NKccSqWM-js:-K7Dn4G0DeS8WiOWixCd:1261992712) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L126)):*
In my testing, these border cases seemed to be handled automatically.  I'll look into a more appropriate conflict type.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/69)*

regnrand:
Reviewed 4 of 4 files at r2.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 130 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K7DwvUp1CUSkIvXcF17:-K7DwvUq6J_4Jj--oxhs:-215837674) ([raw file](https://github.com/sillsdev/flexbridge/blob/8b03ad6d2f39555231fd6306e68c63e189d82ac2/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L130)):*
Just throw an exception, since the user can't do S/R with the dict config files but without the schema.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 28 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6BNah62vfMFYOyk17Z:-K7DuNoPOqRoyWf2AY9d:2144559082) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L28)):*
*Much* less code now and easier to read! Thx.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 126 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6B-i1s7NKccSqWM-js:-K7DoNvvFoFZ9QLW5h7H:1335927355) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L126)):*
Trouble is, there isn't only one conflict class type, even for an atomic merge like you want. How about I work up that code and email it to you? It will not be a lot and it will work with Chorus' ethos.

---

*[src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs, line 146 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6AuXt9KUjiUYlJZqHz:-K7DFnQlbOoyq8TZozUz:940195681) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs#L146)):*
True, it wasn't used locally, but the idea of the line was to ensure a brand new  instance of the MDC for each test (test fixture). Who knows, when it may become important? It is probably best to leave it in.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/69)*

papeh:

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 126 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6B-i1s7NKccSqWM-js:-K7HiW7G6JteZ4CuX8Ni:-1374242615) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L126)):*
Thanks. Conflict reports look "better," if you feel that way about more details and full context :-)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/69)*

regnrand:
The PR LGTM now. Thanks for doing this.

---

Reviewed 5 of 5 files at r3.
Review status: 9 of 11 files reviewed at latest revision, 2 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 126 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K6B-i1s7NKccSqWM-js:-K7HtzlyN26BPgYxvX_5:-237842782) ([raw file](https://github.com/sillsdev/flexbridge/blob/b2e8538cbb60d3a0c5ce8a0c666038e5879a183b/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L126)):*
Well, I think the conflict report details tend to stink something awful. ;-)

---

*[src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs, line 168 \[r3\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-K7Hv-YdYmsXSzr72yPs:-K7Hv-YdYmsXSzr72yPt:1356294839) ([raw file](https://github.com/sillsdev/flexbridge/blob/853f66b147799a845761a281125a0be857bf8ac9/src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs#L168)):*
I would 'never' delete one of those. ;-)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/69#-:-K7HvDOcZNBIgI2Ow6t5:-1074827041)*

PR #78, opened 2016-01-27
LT-16969 second sync crash

gmartin7:
* Make sure Temp folder exists
* After copying xsd file, make sure it's not read-only
  (which would interfere with copying it again later)

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/78)


papeh:
:+1:

PR #79, opened 2016-01-28
Update Linux packaging

neilmayhew:
The main purpose is to switch to using automatic dependency generation, but a few other things are tidied up in the process. In particular, the package now builds successfully for 16.04 (Xenial).

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/79)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,7 +26,7 @@ clean:
 	. ./environ && cd build && xbuild FLExBridge.build.mono.proj /t:Clean /p:RootDir=..
 	/bin/rm -rf output Download Mercurial
 
-install: release
+install:
 	/usr/bin/install -d $(DESTDIR)/usr/lib/flexbridge
 	/usr/bin/install output/ReleaseMono/*.* $(DESTDIR)/usr/lib/flexbridge
 	/bin/chmod -x $(DESTDIR)/usr/lib/flexbridge/*.htm
--- a/build/FLExBridge.build.mono.proj
+++ b/build/FLExBridge.build.mono.proj
@@ -79,6 +79,7 @@
 	<!-- NDeskDbus is required only so that Palaso can be localized on linux -->
 	<NDeskDBusFiles Include="$(RootDir)/lib/$(Configuration)/NDesk.DBus.dll*"/>
 	<ChorusHubFiles Include="$(RootDir)/lib/$(Configuration)/ChorusHub.*"/>
+	<ConfigFiles Include="$(RootDir)/lib/$(Configuration)/*.dll.config"/>
   </ItemGroup>
 
   <Target Name="CopyExtraFilesToOutput">
@@ -88,6 +89,7 @@
 	<Copy SourceFiles="@(LocalizeFiles)" DestinationFolder="$(RootDir)/output/$(Configuration)/localizations"/>
 	<Copy SourceFiles="@(NDeskDBusFiles)" DestinationFolder="$(RootDir)/output/$(Configuration)"/>
 	<Copy SourceFiles="@(ChorusHubFiles)" DestinationFolder="$(RootDir)/output/$(Configuration)"/>
+	<Copy SourceFiles="@(ConfigFiles)" DestinationFolder="$(RootDir)/output/$(Configuration)"/>
   </Target>
 
   <Target Name="Compile" DependsOnTargets="CopyExtraFilesToOutput; RestorePackages">
--- a/build/NuGet.targets
+++ b/build/NuGet.targets
@@ -49,7 +49,7 @@
 	<Target Name="_DownloadNuGet" Condition=" '$(DownloadNuGetExe)' == 'true' AND !Exists('$(NuGetExePath)')">
 		<DownloadNuGet OutputFilename="$(NuGetExePath)"
 			Condition="'$(OS)' == 'Windows_NT'" />
-		<Exec Command="wget https://nuget.org/NuGet.exe || curl -O -L wget https://nuget.org/NuGet.exe"
+		<Exec Command="wget https://nuget.org/NuGet.exe || curl -O -L https://nuget.org/NuGet.exe"
 			WorkingDirectory="$(NuGetToolsPath)"
 			Condition="'$(OS)' != 'Windows_NT'" />
 	</Target>
--- a/debian/control
+++ b/debian/control
@@ -3,13 +3,10 @@ Section: misc
 Priority: extra
 Maintainer: Stephen McConnel <stephen_mcconnel@sil.org>
 Build-Depends: debhelper (>= 8.0.0),
+ cli-common-dev (>= 0.8~),
  mono-sil, libgdiplus-sil,
- cli-common-dev,
- geckofx29, unzip, wget,
- pkg-config,
- curl,
- ca-certificates,
- adduser (>= 3.11)
+ libenchant-dev, libxklavier-dev, libgtk2.0-dev,
+ geckofx29, unzip, curl, ca-certificates
 Standards-Version: 3.9.2
 Homepage: http://projects.palaso.org/projects/fwbridge
 #Vcs-Git: git://git.debian.org/collab-maint/flexbridge.git
@@ -17,10 +14,11 @@ Homepage: http://projects.palaso.org/projects/fwbridge
 
 Package: flexbridge
 Architecture: any
-Depends: ${shlibs:Depends}, ${misc:Depends}, python (>= 2.7), python (<< 2.8),
- mono-sil, libgdiplus-sil, libgtk2.0-dev, libgtkmm-2.4-1c2a, libgtkmm-2.4-dev,
- libgtksourceview2.0-0, libgtksourceview2.0-common, geckofx29, unzip,
- fieldworks-applications (>> 8.1.9)
+Depends: ${shlibs:Depends}, ${misc:Depends}, ${cli:Depends},
+ mono-sil, libgdiplus-sil,
+ fieldworks-applications (>> 8.1.9),
+ python (>= 2.7), python (<< 2.8),
+ geckofx29, unzip, adduser (>= 3.11)
 Description: Allow multiple FieldWorks users to collaborate remotely
  FLEx Bridge is an application that allows multiple FieldWorks 8.0 users
  to do remote collaboration (i.e., not necessarily connected by a local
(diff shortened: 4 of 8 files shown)


marksvc:
Really, so it's now an architecture independent package. That's great. I'm so used to not being able to do that :)

---

Reviewed 8 of 8 files at r1.

---

*[debian/control, line 8 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K983fc0f2cEsM6S7rfc:-K983fc1a8qjYloBOxxk:1040635813) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/control#L8)):*
Are you sure it needs libgtk2.0-dev for building/packaging? And Why have the gtk dependency here and not have something like libgtk2.0-0 as a dependency down in Depends? Maybe the system is smart enough to include libgtk2.0-0 in cli:Depends or misc:Depends?

It also seems odd to be *adding* more to Build-Depends, like libenchant-dev and libxklavier-dev, since the package was building/packaging successfully before. Are these needed?

---

*[debian/rules, line 12 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K985518cx22ybOHpMKO:-K985519DwYBQUlmgt5P:2127257776) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/rules#L12)):*
Okay, so we exclude icu48, but does it not need icu at all? I guess if you grep about in the tree and it's not there then I guess that's okay.

---

*[download_dependencies_linux.sh, line 56 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K9883Fiv5eN_1JcguCn:-K9883Fiv5eN_1JcguCo:1779558508) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/download_dependencies_linux.sh#L56)):*
Running basename on a URL sounds like it might have problems :).

But it will probably work okay as long as no web server decides to use backslashes for path separators.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-:-K98C53iWIcrsMhAyLpS:175046494)*

ermshiperete:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[debian/control, line 9 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K98H17x7__eWkx75PI1:-K98H17x7__eWkx75PI2:33743353) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/control#L9)):*
Just out of curiosity: why use curl instead of wget?

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/79)*

neilmayhew:
*[download_dependencies_linux.sh, line 56 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K9883Fiv5eN_1JcguCn:-K98JkMSmOV3VTPzlFds:21098329) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/download_dependencies_linux.sh#L56)):*
The entire file is auto-generated by `BuildUpdate`, so you'd have to take that up with Chris :-)

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/79)*

neilmayhew:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[debian/control, line 8 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K983fc0f2cEsM6S7rfc:-K98Glu7Wdmls4XpQlow:121691263) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/control#L8)):*
The reason for adding these to the build depends rather than the runtime depends is so that the runtime depends can be generated automatically. After inspecting the assemblies for `DllImports`, `dh_clideps` needs to know which package provides that shared library, and the package needs to be installed in order for `dh_clideps` to have access to that information. So, you can build the package without those libraries being present, except that you can't do the final step which is to generate the dependencies. Also, it's better to use the more generic name of the `-dev` package and let `dh_clideps` figure out the specific name of the shared library package (eg as with `libgtkmm-2.4-1c2a`).

---

*[debian/control, line 9 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K98H17x7__eWkx75PI1:-K98ImlvZxsp1C4_BIJl:-1265339399) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/control#L9)):*
It's slightly smaller, and Chris prefers it for `BuildUpdate` (IOW, he tests it better :-) ). `BuildUpdate` is what generates the scripts for downloading dependencies.

---

*[debian/rules, line 12 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K985518cx22ybOHpMKO:-K98J9Hh0PApncT32Sru:1936834216) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/rules#L12)):*
It can't have been using it before, because there is no ICU 4.8 available, and it doesn't appear as a dependency. So I assume FB never takes that code path, and at least this makes the situation no worse than it was before. If we don't exclude these module refs, the whole thing blows up because they don't exist.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/79)*

neilmayhew:
*[debian/rules, line 12 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K985518cx22ybOHpMKO:-K98KMKnuY0KYlgE0GqF:27135486) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/rules#L12)):*
I guess it's possible ICU 4.8 is coming from FW, although I thought the version there was upgraded some time ago. However, I'm a bit out of touch with what's being used outside of `develop`.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/79)*

neilmayhew:
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[debian/control, line 9 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K98H17x7__eWkx75PI1:-K98K6JZ5In2x_Mkinn7:1566878828) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/control#L9)):*
Previously, both `wget` and `curl` were listed as dependencies, which is quite unnecessary, although until the typo in `NuGet.targets` was fixed, they were in fact both necessary.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/79)*

ermshiperete:

---

*[debian/rules, line 12 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K985518cx22ybOHpMKO:-K98HXQrYLjgowrG1og8:-1753677178) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/rules#L12)):*
It's coming from libpalaso or icu.net.dll. FW uses ICU 5.4.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/79)*

ermshiperete:
Reviewed 8 of 8 files at r1.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/79)*

marksvc:
LGTM

---

---

*[debian/control, line 8 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K983fc0f2cEsM6S7rfc:-K98MbSqjBZHGCU2SjFo:-1362554339) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/control#L8)):*
Interesting; thank you.

---

*[download_dependencies_linux.sh, line 56 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K9883Fiv5eN_1JcguCn:-K98N5a_CHv8rgXZ05ep:321438895) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/download_dependencies_linux.sh#L56)):*
Oh, and here I assumed I was reviewing some bash code you wrote. Makes sense now given that I never saw a 'source' change for where the bash was coming from.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-:-K98QNRrtpJBl97ZZNLe:676834233)*

neilmayhew:
Thanks!

---

---

*[debian/rules, line 12 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-K985518cx22ybOHpMKO:-K9EtnLAUWvoV-NxSav8:541976311) ([raw file](https://github.com/sillsdev/flexbridge/blob/090704f840b0d55e8c04b16d64a4eff82750ca89/debian/rules#L12)):*
The *references* to `icuuc48` and `icuin48` are coming from the version of `icu.net.dll` that Flexbridge downloads as a dependency. 

The `flexbridge` binary package doesn't contain any shared libraries, and the system version of ICU (`libicu-dev`) has been at version 5.2 since Trusty and 5.5 since Wily. So I don't think Flexbridge can be calling any ICU code: if it tried, it would crash. `icu.net.dll` is needed by both `Palaso.dll` and `PalasoUIWindowsForms.dll`, but presumably Flexbridge doesn't call any code from those libraries that results in a call to ICU. I assume we could in fact remove `icu.net.dll` from the list of downloaded dependencies, but I wouldn't like to risk this without further investigation.

We need to get to the bottom of this, but this change doesn't alter the current situation in any way so I think we can let it go for now.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/79#-:-K9EtiOyXrgdcV6Uyhyq:-1834867277)*

PR #80, opened 2016-01-28
LT-16969 even better

gmartin7:
* Make sure existing Temp xsd file doesn't get in the way
* Make sure we clean up after ourselves when we're done

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/80)

### DIFF
--- a/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs
+++ b/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs
@@ -134,7 +134,7 @@ namespace FLEx_ChorusPlugin.Infrastructure.ActionHandlers
 				Directory.CreateDirectory(xsdDirInProject);
 			var xsdPathInProject = Path.Combine(xsdDirInProject, SharedConstants.DictConfigSchemaFilename);
 			SafelyDeleteDictConfigXsd(xsdPathInProject);
-			File.Copy(xsdPath, xsdPathInProject, true);
+			File.Copy(xsdPath, xsdPathInProject);
 			return xsdPathInProject;
 		}
 
--- a/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs
+++ b/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs
@@ -36,7 +36,7 @@ namespace FLEx_ChorusPlugin.Infrastructure.ActionHandlers
 		{
 			// -p <$fwroot>\foo\foo.fwdata
 			var projectDir = Path.GetDirectoryName(commandLineArgs["-p"]);
-			EnsureAccessToDictionaryConfigXsd(commandLineArgs);
+			var tempXsdPath = EnsureAccessToDictionaryConfigXsd(commandLineArgs);
 			using (var chorusSystem = Utilities.InitializeChorusSystem(projectDir, commandLineArgs["-u"], FlexFolderSystem.ConfigureChorusProjectFolder))
 			{
 				var newlyCreated = false;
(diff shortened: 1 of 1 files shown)


regnrand:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 137 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/80#-K99DZpNqEjXxN-nCJrR:-K99DZpNqEjXxN-nCJrS:-1641731312) ([raw file](https://github.com/sillsdev/flexbridge/blob/3da65daab10876c3ba8bac3b4c9aa67e2a970658/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L137)):*
1. Seems like you don't need that 'true' now, since you just deleted it.

2. Why not set the attribute to normal after the copy, as well? It isn't safe to leave it read only, is it?

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 144 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/80#-K99EbIgB71fJHuwE7po:-K99EbIhdBgTRYIYO30_:-385492812) ([raw file](https://github.com/sillsdev/flexbridge/blob/3da65daab10876c3ba8bac3b4c9aa67e2a970658/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L144)):*
File.Exists docs say it will return 'false' even if it does exist, but the user doesn't have proper permissions. So, if it has 'wrong' permissions, but exists, it won't be deleted.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/80)*

papeh:
:lgtm: 

---

Reviewed 1 of 1 files at r2.

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 137 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/80#-K99DZpNqEjXxN-nCJrR:-K99PoeO5rDZXcxf3lGq:201574247) ([raw file](https://github.com/sillsdev/flexbridge/blob/3da65daab10876c3ba8bac3b4c9aa67e2a970658/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L137)):*
I'd think it would be safe to leave it read-only, since we're changing the attribute before deleting it.

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 144 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/80#-K99EbIgB71fJHuwE7po:-K99PCX7zqfvmreRTFWB:-245809099) ([raw file](https://github.com/sillsdev/flexbridge/blob/3da65daab10876c3ba8bac3b4c9aa67e2a970658/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L144)):*
I'm assuming in that case we couldn't  delete it anyway and would therefore still crash.

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 116 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/80#-K99QB0rmg4BYMA-kYk5:-K99QB0rmg4BYMA-kYk6:-358844281) ([raw file](https://github.com/sillsdev/flexbridge/blob/e5c223e084bbfd0ef277cf8e891f94753dfc1e26/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L116)):*
this makes what I think is a reasonable assumption: deleting the tempXsdPath won't throw. It's most important that the lockPathname is deleted.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/80#-:-K99QVUyAtqEiLDKQR7y:980077423)*

gmartin7:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs, line 144 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/80#-K99EbIgB71fJHuwE7po:-K99RjuIjM0BTnZyX3bU:-1214343867) ([raw file](https://github.com/sillsdev/flexbridge/blob/3da65daab10876c3ba8bac3b4c9aa67e2a970658/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/SendReceiveActionHandler.cs#L144)):*
File.Exists returns true in the read-only case that we're concerned about. Another case would still crash.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/80)*

PR #82, opened 2016-02-01
Added some more exception info

gmartin7:
* throw an exception if the xsd file is not found
* moved the delete down farther in case the xsd file
  was being deleted too soon.
* Log when xsd in Temp folder is deleted

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/82)


regnrand:
I think I've had about as much fun as I can stand with manipulating that FLEx schema in FB code. Make FLEx, not FB, be responsible to put it in a specified location in the project and keeping it updated. Then, FB can just use it, knowing it will be in that location. FB ought not need to worry about finding it, copying it, deleting if, or updating it.

---

Review status: 0 of 2 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/82#-:-K9U1MZa5OS_X81pPEO1:-1224343257)*

PR #85, opened 2016-02-09
Update NDesk.DBus.dll to v0.12

marksvc:
To match the NDesk.DBus.dll used in libpalaso v2.6.

Fixes crash described in LT-17058.
Just manually updating the dll in the flexbridge repository for now.

Also including .dll.config file which is needed to not crash.

=====

I tested using the DLL and .dll.config and it prevents the crash.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/85)


papeh:
seems there should be a better way to do this, but if it works, I suppose it's :ok:. Glad we don't have to support Solaris.

StephenMcConnel:
LGTM

PR #86, opened 2016-02-18
LT-17100: Uninstall FLExBridge before installing

papeh:
When the first three version numbers and the Upgrade Code match,
but the Prouct Code does not, MSI allows installing both versions
as "separate products." This is not good. AllowSameVersionUpgrades
interprets installing these "same versions" as an "upgrade,"
therefore properly removing the installed version before installing.

[Review on Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/86)

### DIFF
--- a/src/Installer/Installer.wxs
+++ b/src/Installer/Installer.wxs
@@ -26,7 +26,7 @@ http://blogs.msdn.com/robmen/archive/2003/10/04/56479.aspx -->
 	<WixVariable Id="WixUIDialogBmp" Value="DialogBGWithSILLogo.bmp" />
 
 	<!-- Using afterInstallFinalize will allow the apparent downgrading of files in the Chorus merge module. -->
-	<MajorUpgrade AllowDowngrades="no" DowngradeErrorMessage="A newer version of FLEx Bridge is already installed. Setup will now exit." Schedule="afterInstallValidate"/>
+	<MajorUpgrade AllowSameVersionUpgrades="yes" AllowDowngrades="no" DowngradeErrorMessage="A newer version of FLEx Bridge is already installed. Setup will now exit." Schedule="afterInstallValidate"/>
 
 	<Property Id="MANUFACTURERREGKEY">SIL</Property>
 	<!-- The following will set properties A[FW8INSTALLDIR] and B[FW8DEVINSTALLDIR] to the result of two different registry searches.


regnrand:
Reviewed 1 of 1 files at r1.

---

*[src/Installer/Installer.wxs, line 29 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/86#-KAq6cTZQvxfc5QwNaf6:-KAq6cTZQvxfc5QwNaf7:-525860823) ([raw file](https://github.com/sillsdev/flexbridge/blob/458fa7ef8687a568824e782619703fd21fbc37e5/src/Installer/Installer.wxs#L29)):*
The core problem goes back to when you all decided to do installers as continuous builds on Team City without changing the version number. Get rid of continuous build installers, and you won't need this hack.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/86)*

papeh:

---

*[src/Installer/Installer.wxs, line 29 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/86#-KAq6cTZQvxfc5QwNaf6:-KAq92KeiAMfJx-CDFHC:-324967708) ([raw file](https://github.com/sillsdev/flexbridge/blob/458fa7ef8687a568824e782619703fd21fbc37e5/src/Installer/Installer.wxs#L29)):*
I don't think we can do without at least nightly installers for our test team. I may be able to convince @Jason to include a ticking 4th-place version number, but  as long as the first three are the same, we need this.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/86)*

regnrand:

---

*[src/Installer/Installer.wxs, line 29 \[r1\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/86#-KAq6cTZQvxfc5QwNaf6:-KAqApbiGZXdeGe1_MV0:-567220116) ([raw file](https://github.com/sillsdev/flexbridge/blob/458fa7ef8687a568824e782619703fd21fbc37e5/src/Installer/Installer.wxs#L29)):*
Installers ignore that fourth digit.

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/sillsdev/flexbridge/86)*

PR #101, opened 2016-06-28
Update working directory after pulling changes

ermshiperete:
This fixes the problem in LfMergeBridge where changes we got from
others didn't show up in LanguageForge because the working directory
didn't get updated.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/101)

### DIFF
--- a/src/LfMergeBridge/LanguageForgeSendReceiveActionHandler.cs
+++ b/src/LfMergeBridge/LanguageForgeSendReceiveActionHandler.cs
@@ -93,7 +93,8 @@ namespace LfMergeBridge
 			// Do a pull first, to see if FLEx user has upgraded.
 			var uri = options[LfMergeBridgeUtilities.languageDepotRepoUri];
 			var repositoryAddress = RepositoryAddress.Create(options[LfMergeBridgeUtilities.languageDepotRepoName], uri, false);
-			if (hgRepository.Pull(repositoryAddress, uri))
+			var pulledChangesFromOthers = hgRepository.Pull(repositoryAddress, uri);
+			if (pulledChangesFromOthers)
 			{
 				// Check for a higher branch that came in.
 				var highestHead = LfMergeBridgeUtilities.GetHighestRevision(hgRepository);
@@ -131,8 +132,18 @@ namespace LfMergeBridge
 				return;
 			}
 
+			if (pulledChangesFromOthers)
+			{
+				// ENHANCE: A better fix would be in Chorus. Chorus should notice there was no new
+				// commit on the local branch, but that there is a higher head of the same branch
+				// than that of the current working set, and thus, it should tell the adjunct to
+				// do its simple update.
+				hgRepository.UpdateToBranchHead(desiredBranchName);
+				syncAdjunct.SimpleUpdate(progress, false);
+			}
+
 			// Fwdata file has been restored by this point.
-			var gotChangesText = syncResults.DidGetChangesFromOthers ? "Received changes from others" : "No changes from others";
+			var gotChangesText = (pulledChangesFromOthers || syncResults.DidGetChangesFromOthers) ? "Received changes from others" : "No changes from others";
 			// LF Merge needs to know if anything came from LD. Since new stuff did come in, then LF has to rebuild its FdoCache.
 			LfMergeBridgeUtilities.AppendLineToSomethingForClient(ref somethingForClient, string.Format("{0} {1}: {2}", syncBase, LfMergeBridgeUtilities.success, gotChangesText));
 			progress.WriteVerbose(gotChangesText);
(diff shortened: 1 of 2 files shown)


ermshiperete:
Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/LfMergeBridge/LanguageForgeSendReceiveActionHandler.cs, line 138 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/101#-KLXE07r-ZmJAqHt_oTX:-KLXE07sDQqdam1XrNz5:1915398706) ([raw file](https://github.com/sillsdev/flexbridge/blob/000ff7867562646bca4d468a39833905d957c8c0/src/LfMergeBridge/LanguageForgeSendReceiveActionHandler.cs#L138)):*
Randy's comments on Skype:

> In an ideal world, lfmerge would have just asked for updates (new type of handler), and skipped a fuller S/R, since I assume it knew it had made no changes. But, the world isn't ideal, so I guess the change makes sense.
> A better fix would be in Chorus. Chorus should notice there was no new commit on the local branch, but that there is a higher head of the same branch than that of the current working set, and thus, it should tell the adjunct to do its simple update.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/101)*

jasonleenaylor:
The commit comment on your refactoring could use a little adjustment.

jasonleenaylor:
Do the other action handlers do a similar operation? If not then there might be a better fix to be had. If no better solution is obvious I have no major objections to this practical fix.

ermshiperete:
This action handler is unique in that it does a pull first in order to be able to check the highest branch. This is necessary because LfMerge doesn't do data migrations. Because we do a pull outside of Chorus, Chorus doesn't see that there are new changes (unless there were local changes) and so doesn't update the working directory.

The improvement that Randy suggests in Chorus (see comment in `LanguageForgeSendReceiveActionHandler.cs`) might be a better way, but since it needs to be a general solution (instead of the LfMerge targeted one here) it needs careful implementation and unit tests. I think I'd prefer to add a ENHANCE comment and merge it with the workaround for now.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/101#-:-KL_Ep8FIn0dIduD49ZZ:-997176688)*

rmunn:
As I said below, I agree that a practical, get-this-working solution that works NOW is better than a perfect solution that will work later, for two reasons. First, this doesn't preclude going back to enhance the code later; and second, we have users clamoring for the LF Send/Receive feature to be working. So I'm in favor of merging this workaround now to get things working, and coming back later to work on the better approach that Randy suggested.

---

Review status: 0 of 2 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[src/LfMergeBridge/LanguageForgeSendReceiveActionHandler.cs, line 138 \[r2\]](https://reviewable.io:443/reviews/sillsdev/flexbridge/101#-KLXE07r-ZmJAqHt_oTX:-KL_HMtQvX0ojD4e6lAC:1348558643) ([raw file](https://github.com/sillsdev/flexbridge/blob/d36e4b3dd0e4219e854032dc91d4e4a0d74f3fa9/src/LfMergeBridge/LanguageForgeSendReceiveActionHandler.cs#L138)):*
<details><summary><i>Previously, ermshiperete (Eberhard Beilharz) wrote…</i></summary>
> Randy's comments on Skype:
> 
> > In an ideal world, lfmerge would have just asked for updates (new type of handler), and skipped a fuller S/R, since I assume it knew it had made no changes. But, the world isn't ideal, so I guess the change makes sense.
> > A better fix would be in Chorus. Chorus should notice there was no new commit on the local branch, but that there is a higher head of the same branch than that of the current working set, and thus, it should tell the adjunct to do its simple update.
> 
</details>
I agree that an update handler would be a good thing to have, and perhaps we can do that in a future version where we try to optimize S/R speed in LfMerge. Right now, though, our priority needs to be getting LfMerge into a working state so we can let users start using it. So a solution that isn't perfect, but works, is good enough for this release.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/101#-:-KL_KQIL-FWJl3ahtqRm:28071794)*

rmunn:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/101)*

ermshiperete:
Reviewed 1 of 2 files at r1, 1 of 1 files at r4.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/101)*

rmunn:
:lgtm:

---

Reviewed 1 of 1 files at r4.
Review status: all files reviewed at latest revision, all discussions resolved, some commit checks pending.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/101#-:-KL_MhQ6ivS_jTz2hTeg:1417088721)*

PR #102, opened 2016-08-18
Allow to set username when cloning

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/102)


jasonleenaylor:
I don't necessarily understand the need, but the change looks reasonable.

ermshiperete:
The change is necessary because for LanguageForge S/R we want a sensible username to show up instead of the name of the user that runs LanguageForge (which is "www-data").

PR #104, opened 2016-08-19
Include all architectures of mercurial

marksvc:
flexbridge is packaged as an any architecture package. So since it
bundles Mercurial, it needs to bundle both Mercurial architectures so
it can run on machines of either architecture.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/104)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -35,7 +35,9 @@ install: release
 	/bin/chmod -x $(DESTDIR)/usr/lib/flexbridge/*.md*
 	/usr/bin/install lib/common/setup-user.sh $(DESTDIR)/usr/lib/flexbridge
 	/usr/bin/install lib/common/run-app $(DESTDIR)/usr/lib/flexbridge
-	/usr/bin/install -m644 lib/ReleaseMono/Mercurial-$(CPUARCH).zip $(DESTDIR)/usr/lib/flexbridge
+	# Copy mercurial for both architectures since flexbridge is an any architecture package.
+	/usr/bin/install -m644 lib/ReleaseMono/Mercurial-x86_64.zip $(DESTDIR)/usr/lib/flexbridge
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -35,7 +35,9 @@ install: release
 	/bin/chmod -x $(DESTDIR)/usr/lib/flexbridge/*.md*
 	/usr/bin/install lib/common/setup-user.sh $(DESTDIR)/usr/lib/flexbridge
 	/usr/bin/install lib/common/run-app $(DESTDIR)/usr/lib/flexbridge
-	/usr/bin/install -m644 lib/ReleaseMono/Mercurial-$(CPUARCH).zip $(DESTDIR)/usr/lib/flexbridge
+	# Copy mercurial for both architectures since flexbridge is an any architecture package.
```
ermshiperete: Wouldn't it be better to make the package architecture dependent instead of "any"?

marksvc: That may be. FB was *just* changed to be an architecture-independent package earlier this year, IIRC. But we could go to the work to change it back.

jasonleenaylor:
How does this interact with the post install unzip?

marksvc:
Jason, so at https://github.com/sillsdev/flexbridge/blob/develop/debian/postinst#L26 it should unzip just the zip file for the current machine's architecture. I could test by making and installing the packages using my local packaging system if anyone's concerned that it might not work.

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/104#-:-KPmznu2dqg0X23N46GN:bf301e7)*

jasonleenaylor:
:lgtm:

---

Review status: 0 of 1 files reviewed at latest revision, 1 unresolved discussion.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/104#-:-KPngrRqXfigcm-V-Dv2:bnfp4nl)*

PR #116, opened 2016-09-29
Fix FB model to match FW model; Reinforce Unit Tests

papeh:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/116)

### DIFF
--- a/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs
+++ b/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs
@@ -4,12 +4,16 @@
 // Distributable under the terms of the MIT License, as specified in the license.rtf file.
 // --------------------------------------------------------------------------------------------
 
+using System;
 using System.IO;
 using System.Reflection;
+using System.Threading;
 using FLEx_ChorusPlugin.Infrastructure;
+using FLEx_ChorusPlugin.Infrastructure.Handling;
 using FLEx_ChorusPlugin.Infrastructure.Handling.ConfigLayout;
 using NUnit.Framework;
 using Palaso.IO;
+using Palaso.Progress;
 using Palaso.TestUtilities;
 using TriboroughBridge_ChorusPlugin;
 
@@ -20,6 +24,7 @@ namespace FLEx_ChorusPluginTests.Infrastructure.Handling.ConfigLayout
 	{
 		private TempFile _configFile;
 		private TemporaryFolder _tempFolder;
+		private string _xsdPathInProjRepo;
 
 		[TestFixtureSetUp]
 		public override void FixtureSetup()
@@ -29,16 +34,17 @@ namespace FLEx_ChorusPluginTests.Infrastructure.Handling.ConfigLayout
 			var appsDir = Path.GetDirectoryName(Utilities.StripFilePrefix(Assembly.GetExecutingAssembly().CodeBase));
 			var xsdPath = Path.Combine(appsDir, "TestData", "Language Explorer", "Configuration", SharedConstants.DictConfigSchemaFilename);
 			_tempFolder = new TemporaryFolder("Temp");
-			var xsdPathInProj = Path.Combine(_tempFolder.Path, SharedConstants.DictConfigSchemaFilename);
-			File.Copy(xsdPath, xsdPathInProj, true);
+			_xsdPathInProjRepo = Path.Combine(_tempFolder.Path, SharedConstants.DictConfigSchemaFilename);
+			File.Copy(xsdPath, _xsdPathInProjRepo, true);
 		}
 
 		[TestFixtureTearDown]
 		public override void FixtureTearDown()
 		{
-			base.FixtureTearDown();
+			ClearCachedSchemaPathFromHandler();
 			_tempFolder.Dispose();
 			_tempFolder = null;
+			base.FixtureTearDown();
 		}
 
 		[SetUp]
@@ -51,21 +57,50 @@ namespace FLEx_ChorusPluginTests.Infrastructure.Handling.ConfigLayout
 		[TearDown]
 		public override void TestTearDown()
 		{
-			base.TestTearDown();
 			_configFile.Dispose();
 			_configFile = null;
+			base.TestTearDown();
 		}
 		
 		[Test]
 		public void CanValidateFiles() // Config File and Schema both exist
 		{
-			Assert.That(new DictionaryConfigurationHandlerStrategy().CanValidateFile(_configFile.Path));
+			RunOneAtATime(() =>Assert.That(FileHandler.CanValidateFile(_configFile.Path), "Should be able to validate against the schema"));
 		}
 
 		[Test]
-		public void ValidatesValidFile()
+		public void CantValidateOtherFiles() // Config File is of the wrong type
 		{
-			const string configXml = @"<?xml version='1.0' encoding='utf-8'?>
+			RunOneAtATime(() =>
+			{
+				using (var configFile = TempFile.WithExtension(".incorrect"))
+					Assert.False(FileHandler.CanValidateFile(configFile.Path));
+			});
+		}
+
+		[Test]
+		public void CantValidateNonexistentFiles() // Config File does not exist
+		{
+			RunOneAtATime(() =>
+			{
+				string path;
+				using (var configFile = TempFile.WithExtension(SharedConstants.fwdictconfig))
+					path = configFile.Path;
+				Assert.False(FileHandler.CanValidateFile(path));
+			});
+		}
+
+		[Test]
+		public void CantValidateWithoutSchema() // Schema does not exist
+		{
+			RunOneAtATime(() =>
+			{
+				using (new SchemaSuppressor(_xsdPathInProjRepo))
+					Assert.False(FileHandler.CanValidateFile(_configFile.Path), "Should not be able to validate without the schema");
+			});
+		}
+
+			const string ValidConfigXml = @"<?xml version='1.0' encoding='utf-8'?>
 <DictionaryConfiguration name='Root-based (complex forms as subentries)' allPublications='true' version='1' lastModified='2014-10-07'>
   <ConfigurationItem name='Main Entry' style='Dictionary-Normal' isEnabled='true' field='LexEntry' cssClassNameOverride='entry'>
   <ParagraphOptions paragraphStyle='Dictionary-Normal' continuationParagraphStyle='Dictionary-Continuation' />
@@ -82,28 +117,90 @@ namespace FLEx_ChorusPluginTests.Infrastructure.Handling.ConfigLayout
 	</ConfigurationItem>
   </ConfigurationItem>
 </DictionaryConfiguration>";
-			File.WriteAllText(_configFile.Path, configXml);
-			Assert.IsNullOrEmpty(new DictionaryConfigurationHandlerStrategy().ValidateFile(_configFile.Path));
+
+		/// <summary>
+		/// Tests the validation code for FW Dict Config files. Data is dated (we do not store a current copy of the schema anywhere in this repo);
+		/// tests only that we locate and use a schema definition when FLEx has put it in the S/R repo in a folder named Temp near the config file.
+		/// </summary>
+		[Test]
+		public void ValidatesValidFile()
+		{
+			RunOneAtATime(() =>
+			{
+				File.WriteAllText(_configFile.Path, ValidConfigXml);
+				Assert.IsNullOrEmpty(FileHandler.ValidateFile(_configFile.Path, new NullProgress()), "Should validate against the schema");
+			});
+		}
+
+		[Test]
+		public void DoesNotValidateWithoutSchema()
+		{
+			RunOneAtATime(() =>
+			{
+				File.WriteAllText(_configFile.Path, ValidConfigXml);
+				using (new SchemaSuppressor(_xsdPathInProjRepo))
+					Assert.Throws<ArgumentException>(() => FileHandler.ValidateFile(_configFile.Path, new NullProgress()),
+						"Should not validate w/o schema");
+			});
 		}
 
 		[Test]
 		public void DoesNotValidateMalformedXmlFile()
 		{
-			const string configXml = @"<?xml version='1.0' encoding='utf-8'?>
+			RunOneAtATime(() =>
+			{
+				const string configXml = @"<?xml version='1.0' encoding='utf-8'?>
 <DictionaryConfiguration name='Root-based (complex forms as subentries)' allPublications='true' version='1' lastModified='2014-10-07'>
   </ConfigurationItem>
 </DictionaryConfiguration>";
-			File.WriteAllText(_configFile.Path, configXml);
-			Assert.That(new DictionaryConfigurationHandlerStrategy().ValidateFile(_configFile.Path), Is.Not.Null.Or.Empty);
+				File.WriteAllText(_configFile.Path, configXml);
+				Assert.IsNotEmpty(FileHandler.ValidateFile(_configFile.Path, new NullProgress()));
+			});
 		}
 
 		[Test]
 		public void DoesNotValidateInvalidConfigFile()
 		{
-			const string configXml = @"<?xml version='1.0' encoding='utf-8'?>
+			RunOneAtATime(() =>
+			{
+				const string configXml = @"<?xml version='1.0' encoding='utf-8'?>
 <DictionaryConfiguration name='Root-based (complex forms as subentries)' allPublications='true' version='1' badAttribute='prohibited'/>";
-			File.WriteAllText(_configFile.Path, configXml);
-			Assert.That(new DictionaryConfigurationHandlerStrategy().ValidateFile(_configFile.Path), Is.Not.Null.Or.Empty);
+				File.WriteAllText(_configFile.Path, configXml);
+				Assert.IsNotEmpty(FileHandler.ValidateFile(_configFile.Path, new NullProgress()));
+			});
+		}
+
+		private void RunOneAtATime(Action test)
+		{
+			lock (_tempFolder)
+			{
+				ClearCachedSchemaPathFromHandler();
+				test();
+				while (!File.Exists(_xsdPathInProjRepo))
+					Thread.Sleep(10);
+			}
+		}
+
+		private void ClearCachedSchemaPathFromHandler()
+		{
+			var handler = ((FieldWorksCommonFileHandler)FileHandler).GetHandlerfromExtension(SharedConstants.fwdictconfig);
+			((DictionaryConfigurationHandlerStrategy)handler)._xsdPathname = null;
+		}
+
+		private sealed class SchemaSuppressor : IDisposable
+		{
+			private readonly string _suppressedSchemaPath;
+
+			public SchemaSuppressor(string schemaPath)
+			{
+				_suppressedSchemaPath = Path.ChangeExtension(schemaPath, "hidden");
+				File.Move(schemaPath, _suppressedSchemaPath);
+			}
+
+			public void Dispose()
+			{
+				File.Move(_suppressedSchemaPath, Path.ChangeExtension(_suppressedSchemaPath, "xsd"));
+			}
 		}
 	}
 }
--- a/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs
+++ b/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs
@@ -7,10 +7,7 @@
 using System;
 using System.IO;
 using System.Reflection;
-using System.Threading;
 using FLEx_ChorusPlugin.Infrastructure;
-using FLEx_ChorusPlugin.Infrastructure.Handling;
-using FLEx_ChorusPlugin.Infrastructure.Handling.ConfigLayout;
 using NUnit.Framework;
 using Palaso.IO;
 using Palaso.Progress;
@@ -23,33 +20,25 @@ namespace FLEx_ChorusPluginTests.Infrastructure.Handling.ConfigLayout
 	public class DictionaryConfigurationHandlerStrategyTests : BaseFieldWorksTypeHandlerTests
 	{
 		private TempFile _configFile;
-		private TemporaryFolder _tempFolder;
-		private string _xsdPathInProjRepo;
+		private string _xsdSourcePath;
 
 		[TestFixtureSetUp]
 		public override void FixtureSetup()
 		{
-			base.FixtureSetup();
 			// Copy the schema file to where the Strategy looks
 			var appsDir = Path.GetDirectoryName(Utilities.StripFilePrefix(Assembly.GetExecutingAssembly().CodeBase));
-			var xsdPath = Path.Combine(appsDir, "TestData", "Language Explorer", "Configuration", SharedConstants.DictConfigSchemaFilename);
-			_tempFolder = new TemporaryFolder("Temp");
-			_xsdPathInProjRepo = Path.Combine(_tempFolder.Path, SharedConstants.DictConfigSchemaFilename);
-			File.Copy(xsdPath, _xsdPathInProjRepo, true);
+			_xsdSourcePath = Path.Combine(appsDir, "TestData", "Language Explorer", "Configuration", SharedConstants.DictConfigSchemaFilename);
 		}
 
 		[TestFixtureTearDown]
-		public override void FixtureTearDown()
-		{
-			ClearCachedSchemaPathFromHandler();
-			_tempFolder.Dispose();
-			_tempFolder = null;
-			base.FixtureTearDown();
-		}
+		public override void FixtureTearDown() {} // mask base.FixtureTearDown (it is called by this.TestTearDown)
 
 		[SetUp]
 		public override void TestSetup()
 		{
+			// set up the fixture for each test, so each starts with a fresh HandlerStrategy, because DictionaryConfigurationHandlerStrategy
+			// caches the path to the schema, but some tests require the schema's presence and others require its absence.
+			base.FixtureSetup();
 			base.TestSetup();
 			_configFile = TempFile.WithExtension("." + SharedConstants.fwdictconfig);
 		}
@@ -60,47 +49,41 @@ namespace FLEx_ChorusPluginTests.Infrastructure.Handling.ConfigLayout
 			_configFile.Dispose();
 			_configFile = null;
 			base.TestTearDown();
+			base.FixtureTearDown();
 		}
 		
 		[Test]
 		public void CanValidateFiles() // Config File and Schema both exist
 		{
-			RunOneAtATime(() =>Assert.That(FileHandler.CanValidateFile(_configFile.Path), "Should be able to validate against the schema"));
+			using (TempFolderWithSchema())
+				Assert.That(FileHandler.CanValidateFile(_configFile.Path), "Should be able to validate against the schema");
 		}
 
 		[Test]
 		public void CantValidateOtherFiles() // Config File is of the wrong type
 		{
-			RunOneAtATime(() =>
-			{
-				using (var configFile = TempFile.WithExtension(".incorrect"))
-					Assert.False(FileHandler.CanValidateFile(configFile.Path));
-			});
+			using (TempFolderWithSchema())
+			using (var configFile = TempFile.WithExtension(".incorrect"))
+				Assert.False(FileHandler.CanValidateFile(configFile.Path));
 		}
 
 		[Test]
 		public void CantValidateNonexistentFiles() // Config File does not exist
 		{
-			RunOneAtATime(() =>
-			{
-				string path;
-				using (var configFile = TempFile.WithExtension(SharedConstants.fwdictconfig))
-					path = configFile.Path;
+			string path;
+			using (var configFile = TempFile.WithExtension(SharedConstants.fwdictconfig))
+				path = configFile.Path;
+			using (TempFolderWithSchema())
 				Assert.False(FileHandler.CanValidateFile(path));
-			});
 		}
 
 		[Test]
 		public void CantValidateWithoutSchema() // Schema does not exist
 		{
-			RunOneAtATime(() =>
-			{
-				using (new SchemaSuppressor(_xsdPathInProjRepo))
-					Assert.False(FileHandler.CanValidateFile(_configFile.Path), "Should not be able to validate without the schema");
-			});
+			Assert.False(FileHandler.CanValidateFile(_configFile.Path), "Should not be able to validate without the schema");
 		}
 
-			const string ValidConfigXml = @"<?xml version='1.0' encoding='utf-8'?>
+		private const string ValidConfigXml = @"<?xml version='1.0' encoding='utf-8'?>
 <DictionaryConfiguration name='Root-based (complex forms as subentries)' allPublications='true' version='1' lastModified='2014-10-07'>
   <ConfigurationItem name='Main Entry' style='Dictionary-Normal' isEnabled='true' field='LexEntry' cssClassNameOverride='entry'>
   <ParagraphOptions paragraphStyle='Dictionary-Normal' continuationParagraphStyle='Dictionary-Continuation' />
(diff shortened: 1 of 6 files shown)


regnrand:
Reviewed 1 of 2 files at r1.
Review status: 1 of 2 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs, line 278 at r1](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KSrsX9iMwwtOLsy-HiO:-KSrsX9iMwwtOLsy-HiP:b5ens6p) ([raw file](https://github.com/sillsdev/flexbridge/blob/fd413706b0242ce86ac80aca806edfc82ef02920/src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs#L278)):*
> ```Smalltalk
> 			CheckClassDoesNotExistBeforeUpGrade(mdc, "LexExtendedNote");
> 			CheckPropertyDoesNotExistBeforeUpGrade(mdc, "CmPicture", "DoNotPublishIn");
> 			CheckPropertyDoesNotExistBeforeUpGrade(mdc, "LexPronunciation", "DoNotPublishIn");
> ```

It would be better to check that properties that are to be removed or that have their data types changed actually exist beforehand and that they have the original expected data type.

For example:

1. Make sure that LexSense has a Restrictions property of data type MultiString (et al) This will require a new test method that can test the original data type. The code never had that level of testing before, since we never changed the data type before this DM.
2. Make sure LexEtymology has its Source property of type Unicode.

Also please check all other brand new properties before the migration to make sure they do not exist ahead of time.

Doing all of these pre- and post- checks are essential to know for certain that the expected new model state is a result of the current DM, and not some prior bug.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

papeh:
Review status: 1 of 4 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs, line 278 at r1](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KSrsX9iMwwtOLsy-HiO:-KTAQjwFqXtBR6wf1uz4:b-896fix) ([raw file](https://github.com/sillsdev/flexbridge/blob/fd413706b0242ce86ac80aca806edfc82ef02920/src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs#L278)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> It would be better to check that properties that are to be removed or that have their data types changed actually exist beforehand and that they have the original expected data type.
> 
> For example:
> 
> 1. Make sure that LexSense has a Restrictions property of data type MultiString (et al) This will require a new test method that can test the original data type. The code never had that level of testing before, since we never changed the data type before this DM.
> 2. Make sure LexEtymology has its Source property of type Unicode.
> 
> Also please check all other brand new properties before the migration to make sure they do not exist ahead of time.
> 
> Doing all of these pre- and post- checks are essential to know for certain that the expected new model state is a result of the current DM, and not some prior bug.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

regnrand:
Reviewed 1 of 3 files at r2, 2 of 3 files at r3.
Review status: 3 of 4 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 21 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTATAWMEMdeQRMkIotb:-KTATAWNvMtUqePVyZwD:b-uq320w) ([raw file](https://github.com/sillsdev/flexbridge/blob/31d8f97cab897934dbb2df1a8ada1fad39fa7a80/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L21)):*
> ```Smalltalk
> 	/// Tests the validation code for FW Dict Config files. Data is dated (we do not store a current copy of the schema anywhere in this repo);
> 	/// this tests only that we locate and use a schema definition when FLEx has put it in the S/R repo in a folder named Temp near the config file.
> 	/// </summary>
> ```

Doesn't this comment only apply to the "ValidatesValidFile" test?

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 66 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTATS3_sX0ubYa8A8LZ:-KTATS3_sX0ubYa8A8L_:b-niwc5x) ([raw file](https://github.com/sillsdev/flexbridge/blob/31d8f97cab897934dbb2df1a8ada1fad39fa7a80/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L66)):*
> new DictionaryConfigurationHandlerStrategy()

Replace constructor use in all cases with "FileHandler". You will also need to add a second parameter to places it calls "ValidateFile". The preferred second parm is: new NullProgress().

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 90 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTAX0R-hSafm-WGseGl:-KTAX0R-hSafm-WGseGm:bplaew2) ([raw file](https://github.com/sillsdev/flexbridge/blob/31d8f97cab897934dbb2df1a8ada1fad39fa7a80/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L90)):*
> ```Smalltalk
> </DictionaryConfiguration>";
> 			File.WriteAllText(_configFile.Path, configXml);
> 			Assert.IsNullOrEmpty(new DictionaryConfigurationHandlerStrategy().ValidateFile(_configFile.Path));
> ```

You are probably better off calling "CanValidateFile", rather than "ValidateFile", as "CanValidateFile" better fits with the ethos of not using current data and schema, but only seeing if the schema is present. CanValidate does this:

return FileUtils.CheckValidPathname(pathToFile, Extension) && File.Exists(GetXsdPathname(pathToFile));

You can then test both return options of CanValidate by:

1. Make sure the given "pathToFile" in various options, including testing a mis-matched extension, and file does not existing/not existing. This would require the presence of the schema.

2. Then you should be able to have a couple of tests that have a legitimate data file, where one test does not have the schema and another that does.

---

*[src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs, line 288 at r2](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KSwuJ196JKUlFMFX_cR:-KSwuJ196JKUlFMFX_cS:b-wfa11w) ([raw file](https://github.com/sillsdev/flexbridge/blob/701c076c46336c835601d8975f0d5a57e600ec5d/src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs#L288)):*
> // Make sure merged file does *not* have Sue's changes

---

*[src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs, line 291 at r2](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KSwrLWOJgi1u_xCg5n6:-KSwrLWOJgi1u_xCg5n7:b-jdoppe) ([raw file](https://github.com/sillsdev/flexbridge/blob/701c076c46336c835601d8975f0d5a57e600ec5d/src/FLEx-ChorusPluginTests/Integration/MergeIntegrationTests.cs#L291)):*
> ```Smalltalk
> 					Assert.AreEqual(1, options.Count, "There should be only one WS Option");
> 					Assert.AreEqual("vernacular", options[0].Attribute("id").Value, "should be default vernacular");
> 					Assert.AreEqual("true", options[0].Attribute("isEnabled").Value, "should be enabled");
> ```

Good. Thx.

---

*[src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs, line 334 at r2](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KSwrW0tO5WR-hLC2aMU:-KSwrW0tO5WR-hLC2aMV:b-p8n8th) ([raw file](https://github.com/sillsdev/flexbridge/blob/701c076c46336c835601d8975f0d5a57e600ec5d/src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs#L334)):*
> ```Smalltalk
> 			CheckNewPropertyAfterUpgrade(mdc.GetClassInfo("LexPronunciation"), "DoNotPublishIn", DataType.ReferenceCollection);
> 		}
> #if RANDY_SENT_THIS_VIA_SKYPE
> ```

Yes, and if you use it, then you won't need that "None" kludge.

---

*[src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs, line 432 at r2](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KSwremHpAtVCQV4vwXL:-KSwremHpAtVCQV4vwXM:bts1l1w) ([raw file](https://github.com/sillsdev/flexbridge/blob/701c076c46336c835601d8975f0d5a57e600ec5d/src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs#L432)):*
> ```Smalltalk
> 
> 		private static void CheckPropertyExistsBeforeUpGrade(MetadataCache mdc, string className, string extantPropName,
> 			DataType dataType = DataType.None)
> ```

Get ride of default bogus value.

---

*[src/FLEx-ChorusPlugin/Infrastructure/DataType.cs, line 15 at r2](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KSwrCI-uZOFXIHWHhz7:-KSwrCI-uZOFXIHWHhz8:b-qokcmg) ([raw file](https://github.com/sillsdev/flexbridge/blob/701c076c46336c835601d8975f0d5a57e600ec5d/src/FLEx-ChorusPlugin/Infrastructure/DataType.cs#L15)):*
> ```Smalltalk
> 	{
> 		/// <summary>TESTING</summary>
> 		None = -1,
> ```

Get rid of "None". There is no such data type in FDO and tests don't need it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

papeh:
Review status: 3 of 4 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 66 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTATS3_sX0ubYa8A8LZ:-KTApJI0olz1ig1EsBZS:b-896fix) ([raw file](https://github.com/sillsdev/flexbridge/blob/31d8f97cab897934dbb2df1a8ada1fad39fa7a80/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L66)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> > new DictionaryConfigurationHandlerStrategy()
> 
> Replace constructor use in all cases with "FileHandler". You will also need to add a second parameter to places it calls "ValidateFile". The preferred second parm is: new NullProgress().
</details>
Done.

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 90 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTAX0R-hSafm-WGseGl:-KTBDA3eFhLclq0rHrWa:bvzvr4f) ([raw file](https://github.com/sillsdev/flexbridge/blob/31d8f97cab897934dbb2df1a8ada1fad39fa7a80/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L90)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> You are probably better off calling "CanValidateFile", rather than "ValidateFile", as "CanValidateFile" better fits with the ethos of not using current data and schema, but only seeing if the schema is present. CanValidate does this:
> 
> return FileUtils.CheckValidPathname(pathToFile, Extension) && File.Exists(GetXsdPathname(pathToFile));
> 
> You can then test both return options of CanValidate by:
> 
> 1. Make sure the given "pathToFile" in various options, including testing a mis-matched extension, and file does not existing/not existing. This would require the presence of the schema.
> 
> 2. Then you should be able to have a couple of tests that have a legitimate data file, where one test does not have the schema and another that does.
</details>
I've added a few more tests for `CanValidateFile`

Unfortunately, testing w/o the schema will be tricky, since we cache the path to the schema. So we could:
- test With the schema first, and get a `FileNotFound` exception if we hide the schema to test without it, or
- test withOut the schema first, and not be able to validate with pathname = `string.Empty`

The only way I see around this is to reflectively clear the cached pathname before each test

---

*[src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs, line 334 at r2](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KSwrW0tO5WR-hLC2aMU:-KTAplGAEhOe4TuyRsvw:b-896fix) ([raw file](https://github.com/sillsdev/flexbridge/blob/701c076c46336c835601d8975f0d5a57e600ec5d/src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs#L334)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> Yes, and if you use it, then you won't need that "None" kludge.
</details>
Done.

---

*[src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs, line 432 at r2](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KSwremHpAtVCQV4vwXL:-KTApoMxAPcyebsQvJcx:b-896fix) ([raw file](https://github.com/sillsdev/flexbridge/blob/701c076c46336c835601d8975f0d5a57e600ec5d/src/FLEx-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs#L432)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> Get ride of default bogus value.
</details>
Done.

---

*[src/FLEx-ChorusPlugin/Infrastructure/DataType.cs, line 15 at r2](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KSwrCI-uZOFXIHWHhz7:-KTApzEpaz4cYiOKDbfE:b-896fix) ([raw file](https://github.com/sillsdev/flexbridge/blob/701c076c46336c835601d8975f0d5a57e600ec5d/src/FLEx-ChorusPlugin/Infrastructure/DataType.cs#L15)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> Get rid of "None". There is no such data type in FDO and tests don't need it.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

papeh:
Review status: 3 of 4 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 21 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTATAWMEMdeQRMkIotb:-KTBFR8gLQaNzBCrHU4w:b-896fix) ([raw file](https://github.com/sillsdev/flexbridge/blob/31d8f97cab897934dbb2df1a8ada1fad39fa7a80/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L21)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> Doesn't this comment only apply to the "ValidatesValidFile" test?
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

regnrand:
Review status: 3 of 4 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 66 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTATS3_sX0ubYa8A8LZ:-KTBUt9JEhJyVhYZTDLA:bl9cw94) ([raw file](https://github.com/sillsdev/flexbridge/blob/31d8f97cab897934dbb2df1a8ada1fad39fa7a80/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L66)):*
<details><summary><i>Previously, papeh wrote…</i></summary>
> Done.
</details>
Huh? The latest Exemplar branch head in git still has four case that use: new DictionaryConfigurationHandlerStrategy()

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 90 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTAX0R-hSafm-WGseGl:-KTBVKxTleH7JU5i1KD7:b-s5kozt) ([raw file](https://github.com/sillsdev/flexbridge/blob/31d8f97cab897934dbb2df1a8ada1fad39fa7a80/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L90)):*
<details><summary><i>Previously, papeh wrote…</i></summary>
> I've added a few more tests for `CanValidateFile`
> 
> Unfortunately, testing w/o the schema will be tricky, since we cache the path to the schema. So we could:
> - test With the schema first, and get a `FileNotFound` exception if we hide the schema to test without it, or
> - test withOut the schema first, and not be able to validate with pathname = `string.Empty`
> 
> The only way I see around this is to reflectively clear the cached pathname before each test
</details>
Your FixtureSetup method copies the schema to the expected location. Skip doing that in FixtureSetup but do that copy (in a new test method) for tests that need to find the schema and don't do the copy for tests that do not expect to see it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

papeh:
Review status: 3 of 6 files reviewed at latest revision, 9 unresolved discussions.

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 66 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTATS3_sX0ubYa8A8LZ:-KTBedek6bW9pfkfllFY:b-duknzi) ([raw file](https://github.com/sillsdev/flexbridge/blob/31d8f97cab897934dbb2df1a8ada1fad39fa7a80/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L66)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> Huh? The latest Exemplar branch head in git still has four case that use: new DictionaryConfigurationHandlerStrategy()
</details>
oops, forgot to `push -f`; should be done *now*

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 90 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTAX0R-hSafm-WGseGl:-KTBeko-rKdqYNfUYB4C:b-qoxq53) ([raw file](https://github.com/sillsdev/flexbridge/blob/31d8f97cab897934dbb2df1a8ada1fad39fa7a80/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L90)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> Your FixtureSetup method copies the schema to the expected location. Skip doing that in FixtureSetup but do that copy (in a new test method) for tests that need to find the schema and don't do the copy for tests that do not expect to see it.
</details>
ok, got it working.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

regnrand:
Review status: 3 of 6 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/FieldWorksCommonFileHandler.cs, line 49 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBhJzqtQGZrz0t293z:-KTBhJzqtQGZrz0t294-:b48my38) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPlugin/Infrastructure/Handling/FieldWorksCommonFileHandler.cs#L49)):*
> ```Smalltalk
> 		}
> 
> 		internal IFieldWorksFileHandler GetHandlerfromExtension(string extension)
> ```

This should not be made internal. Access to the impl is to be only via the interface(s).

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 22 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBfx7llb4PuWKkOC67:-KTBfx7m8cNBJK1wEyKe:b2o7oe8) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L22)):*
> ```Smalltalk
> namespace FLEx_ChorusPlugin.Infrastructure.Handling.ConfigLayout
> {
> 	[Export(typeof(IFieldWorksFileHandler))]
> ```

Bummer. Jason pointed out that it did indeed pass without the export. Seems your Dictionary branch isn't using MEF, which is why I needed the export, but you do not. So, you can get rid of it, until you merge the Dictionary branch into develop. Sorry for the false alarm.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 25 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBgT8yJv-YJ7YSPIBE:-KTBgT8yJv-YJ7YSPIBF:b-oqwkdl) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L25)):*
> ```Smalltalk
> 	internal sealed class DictionaryConfigurationHandlerStrategy : IFieldWorksFileHandler
> 	{
> 		internal string _xsdPathname;
> ```

Hmm. Normally data members are private. I don't think tests need it to be internal.

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 187 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBhkWuzyelmkAwicXw:-KTBhkWv8OHEkpUYn7P3:b90asmz) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L187)):*
> ```Smalltalk
> 		{
> 			var handler = ((FieldWorksCommonFileHandler)FileHandler).GetHandlerfromExtension(SharedConstants.fwdictconfig);
> 			((DictionaryConfigurationHandlerStrategy)handler)._xsdPathname = null;
> ```

Test code does not need to access to that method or to that data member.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

papeh:
Review status: 3 of 6 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 22 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBfx7llb4PuWKkOC67:-KTF4GJ4zHrQ5V-Z5WSC:bia4d49) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L22)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> Bummer. Jason pointed out that it did indeed pass without the export. Seems your Dictionary branch isn't using MEF, which is why I needed the export, but you do not. So, you can get rid of it, until you merge the Dictionary branch into develop. Sorry for the false alarm.
</details>
no harm in exporting, though, is there?

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 187 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBhkWuzyelmkAwicXw:-KTF4e_3wV-iXM2Uf4QD:brknolw) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L187)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> Test code does not need to access to that method or to that data member.
</details>
The cached _xsdPathname survives for at least the entire [TestFixture], so it needs to be reset between each [Test]. FLEx tests use `internal` members where reflection could have been used, although I could reflect here if you think that's better practice. 
Otherwise, I'm out of ideas how to make tests w/ and w/ schema to play nicely together.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

regnrand:
Review status: 3 of 6 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 22 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBfx7llb4PuWKkOC67:-KTF_MzpTO5Ne3ed0JQ_:btl9k5v) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L22)):*
<details><summary><i>Previously, papeh wrote…</i></summary>
> no harm in exporting, though, is there?
</details>
There is no reason to export it at this point. True, it does no harm, but then neither does it do any good. Including it at this point is mis-leading, since one might reasonably think think it is needed now.

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 187 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBhkWuzyelmkAwicXw:-KTF8Gq0JTrNCFZI2N7T:bvhfbzc) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L187)):*
<details><summary><i>Previously, papeh wrote…</i></summary>
> The cached _xsdPathname survives for at least the entire [TestFixture], so it needs to be reset between each [Test]. FLEx tests use `internal` members where reflection could have been used, although I could reflect here if you think that's better practice. 
> Otherwise, I'm out of ideas how to make tests w/ and w/ schema to play nicely together.
</details>
Are you forced to use BaseFieldWorksTypeHandlerTests as the superclass of DictionaryConfigurationHandlerStrategyTests?

For cases where you want to reuse the very same instance of an IFieldWorksFileHandler instance, using BaseFieldWorksTypeHandlerTests as the base class is a good idea. Where you want a fresh instance of the IFieldWorksFileHandler impl, as in your set of tests in DictionaryConfigurationHandlerStrategyTests, then using the BaseFieldWorksTypeHandlerTests  base class seems to be unhelpful. Reusing the instance then heads you off into areas of needing access to data members and/or methods that ought not be available to tests.

So, my suggestion is:

1. Remove BaseFieldWorksTypeHandlerTests as the superclass of DictionaryConfigurationHandlerStrategyTests

2. Remove FixtureSetup & FixtureTearDown methods from DictionaryConfigurationHandlerStrategyTests

3. Copy the code  FixtureSetup & FixtureTearDown from BaseFieldWorksTypeHandlerTests and put it in the respective TestSetup & TestTearDown in DictionaryConfigurationHandlerStrategyTests.

4. Copy the TestSetup & TestTearDown code from BaseFieldWorksTypeHandlerTests and add it to the respective methods in DictionaryConfigurationHandlerStrategyTests. (I would add one additional line of code to the TestSetup method, which does this:

Mdc.UpgradeToVersion(MetadataCache.MaximumModelVersion);

Observation 1: By this point, you have a fresh instance of DictionaryConfigurationHandlerStrategy, and its _xsdPathname is null, so there is no test driven need for it to be anything but private.

5. Tests that need to have the schema present can call a new method that has the old copy schema code in it that was taken out of FixtureSetup.

Observation 2: By this point, the tests that do not want to have the schema present won't have it, since they won't call that copy schema method.

Observation 3: By this point, there is no reason to call FieldWorksCommonFileHandler's "GetHandlerfromExtension" method, so it can be restored to being private. A simple call to its public interface CanValidate method will be adequate for the test's purpose.

Observation 4: All of these can be removed:
1) new SchemaSuppressor class,
2) ClearCachedSchemaPathFromHandler method, and
3) RunOneAtATime method

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

papeh:
Review status: 3 of 4 files reviewed at latest revision, 13 unresolved discussions.

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 187 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBhkWuzyelmkAwicXw:-KTG0ytiHirUUu7bY2--:bna3f8) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L187)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> Are you forced to use BaseFieldWorksTypeHandlerTests as the superclass of DictionaryConfigurationHandlerStrategyTests?
> 
> For cases where you want to reuse the very same instance of an IFieldWorksFileHandler instance, using BaseFieldWorksTypeHandlerTests as the base class is a good idea. Where you want a fresh instance of the IFieldWorksFileHandler impl, as in your set of tests in DictionaryConfigurationHandlerStrategyTests, then using the BaseFieldWorksTypeHandlerTests  base class seems to be unhelpful. Reusing the instance then heads you off into areas of needing access to data members and/or methods that ought not be available to tests.
> 
> So, my suggestion is:
> 
> 1. Remove BaseFieldWorksTypeHandlerTests as the superclass of DictionaryConfigurationHandlerStrategyTests
> 
> 2. Remove FixtureSetup & FixtureTearDown methods from DictionaryConfigurationHandlerStrategyTests
> 
> 3. Copy the code  FixtureSetup & FixtureTearDown from BaseFieldWorksTypeHandlerTests and put it in the respective TestSetup & TestTearDown in DictionaryConfigurationHandlerStrategyTests.
> 
> 4. Copy the TestSetup & TestTearDown code from BaseFieldWorksTypeHandlerTests and add it to the respective methods in DictionaryConfigurationHandlerStrategyTests. (I would add one additional line of code to the TestSetup method, which does this:
> 
> Mdc.UpgradeToVersion(MetadataCache.MaximumModelVersion);
> 
> Observation 1: By this point, you have a fresh instance of DictionaryConfigurationHandlerStrategy, and its _xsdPathname is null, so there is no test driven need for it to be anything but private.
> 
> 5. Tests that need to have the schema present can call a new method that has the old copy schema code in it that was taken out of FixtureSetup.
> 
> Observation 2: By this point, the tests that do not want to have the schema present won't have it, since they won't call that copy schema method.
> 
> Observation 3: By this point, there is no reason to call FieldWorksCommonFileHandler's "GetHandlerfromExtension" method, so it can be restored to being private. A simple call to its public interface CanValidate method will be adequate for the test's purpose.
> 
> Observation 4: All of these can be removed:
> 1) new SchemaSuppressor class,
> 2) ClearCachedSchemaPathFromHandler method, and
> 3) RunOneAtATime method
</details>
Thanks; I understand things better now. I think I'll keep the inheritance, but move the `base.FixtureSetup()` call to `TestSetup`, which should get us a fresh Handler for each test.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/FieldWorksCommonFileHandler.cs, line 49 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBhJzqtQGZrz0t293z:-KTFwLXU94SC2gycKbI5:b-896fix) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPlugin/Infrastructure/Handling/FieldWorksCommonFileHandler.cs#L49)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> This should not be made internal. Access to the impl is to be only via the interface(s).
</details>
Done.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 22 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBfx7llb4PuWKkOC67:-KTFwCnJZM9p-bdV08bO:b-896fix) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L22)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> There is no reason to export it at this point. True, it does no harm, but then neither does it do any good. Including it at this point is mis-leading, since one might reasonably think think it is needed now.
</details>
Done.

---

*[src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs, line 25 at r4](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTBgT8yJv-YJ7YSPIBE:-KTFwEfzM-UszQknpZsH:b-896fix) ([raw file](https://github.com/sillsdev/flexbridge/blob/2d7d1d1068db4b7b9537540e10b719985f99d881/src/FLEx-ChorusPlugin/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategy.cs#L25)):*
<details><summary><i>Previously, regnrand (Randy Regnier) wrote…</i></summary>
> Hmm. Normally data members are private. I don't think tests need it to be internal.
</details>
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs
```diff
@@ -23,33 +20,25 @@ namespace FLEx_ChorusPluginTests.Infrastructure.Handling.ConfigLayout
...
-			base.FixtureTearDown();
-		}
+		public override void FixtureTearDown() {} // mask base.FixtureTearDown (it is called by this.TestTearDown)
 
 		[SetUp]
 		public override void TestSetup()
 		{
+			// set up the fixture for each test, so each starts with a fresh HandlerStrategy, because DictionaryConfigurationHandlerStrategy
+			// caches the path to the schema, but some tests require the schema's presence and others require its absence.
+			base.FixtureSetup();
```
regnrand: Calling the base method of some other method is a pretty stinky code smell. I think I'll try my suggestions from the previous review to see if they work.

regnrand:
Review status: 3 of 4 files reviewed at latest revision, 14 unresolved discussions.

---

*[src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs, line 41 at r5](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-KTGBo4azwcUgWgJ_QDP:-KTGBo4azwcUgWgJ_QDQ:b-kjyjcu) ([raw file](https://github.com/sillsdev/flexbridge/blob/9528e420ce73159056a93fb82fd0abe1436fdebe/src/FLEx-ChorusPluginTests/Infrastructure/Handling/ConfigLayout/DictionaryConfigurationHandlerStrategyTests.cs#L41)):*
> ```Smalltalk
> 			// set up the fixture for each test, so each starts with a fresh HandlerStrategy, because DictionaryConfigurationHandlerStrategy
> 			// caches the path to the schema, but some tests require the schema's presence and others require its absence.
> 			base.FixtureSetup();
> ```

Calling the base method of some other method is a pretty stinky code smell. I think I'll try my suggestions from the previous review to see if they work.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

regnrand:
Reviewed 2 of 3 files at r5, 3 of 3 files at r6.
Review status: all files reviewed at latest revision, 14 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116)*

regnrand:
Thx. LGTM

---

Review status: all files reviewed at latest revision, 14 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/116#-:-KTGdzwjTRRVFCt3EaER:b-h7nxgb)*

PR #137, opened 2016-12-22
Update DateModified when doing a merge

ermshiperete:
This fixes a problem when doing S/R in LanguageForge. LF relies on the DateModified time stamp to know which entries it needs to process. If an entry got modified in FLEx and the same entry but
different field got modified in LF, the S/R merged the changes but set the DateModified field to the newer date (which was from LF). This caused LF not to re-process the entry because it already
processed the entry with that time stamp. This caused the modified data from FLEx to not show up in LF (https://trello.com/c/UoyXj4jP).

This change fixes this problem by setting DateModified to the current time in case of a merge.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/137)

### DIFF
--- a/src/LibFLExBridge-ChorusPlugin/Handling/PreferMostRecentTimePreMerger.cs
+++ b/src/LibFLExBridge-ChorusPlugin/Handling/PreferMostRecentTimePreMerger.cs
@@ -1,8 +1,10 @@
-﻿// Copyright (c) 2010-2016 SIL International
+﻿// Copyright (c) 2010-2017 SIL International
 // This software is licensed under the MIT License (http://opensource.org/licenses/MIT) (See: license.rtf file)
 
 using System;
-using System.Globalization;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
 using System.Xml;
 using Chorus.merge.xml.generic;
 
@@ -24,26 +26,40 @@ namespace LibFLExBridgeChorusPlugin.Handling
 	{
 		void IPremerger.Premerge(IMergeEventListener listener, ref XmlNode ourDateTimeNode, XmlNode theirDateTimeNode, XmlNode ancestorDateTimeNode)
 		{
-			RestoreOriginalIfTimestampIsTheOnlyChange(ancestorDateTimeNode, ourDateTimeNode);
-			RestoreOriginalIfTimestampIsTheOnlyChange(ancestorDateTimeNode, theirDateTimeNode);
+			var ourOnlyTimestampChange = RestoreOriginalIfTimestampIsTheOnlyChange(ancestorDateTimeNode, ourDateTimeNode);
+			var theirOnlyTimestampChange = RestoreOriginalIfTimestampIsTheOnlyChange(ancestorDateTimeNode, theirDateTimeNode);
 
-			var newest = DateTime.MinValue.ToString(CultureInfo.InvariantCulture);
-			newest = GetMostRecentVal(newest, ourDateTimeNode);
-			newest = GetMostRecentVal(newest, theirDateTimeNode);
-			UpdateDateTimeVal(newest, ourDateTimeNode);
-			UpdateDateTimeVal(newest, theirDateTimeNode);
+			DateTime newestDateTime;
+			if (ourOnlyTimestampChange && theirOnlyTimestampChange)
+			{
+				// timestamp was the only change. Use the newer timestamp.
+				newestDateTime = DateTime.MinValue;
+				newestDateTime = GetMostRecentVal(newestDateTime, ourDateTimeNode);
+				newestDateTime = GetMostRecentVal(newestDateTime, theirDateTimeNode);
+			}
+			else
+			{
+				// something else besides the timestamp changed. Set timestamp to current time.
+				newestDateTime = DateTime.UtcNow;
+			}
+			var newestDateTimeString = newestDateTime.ToString("yyyy-M-d H:m:s.FFF");
+			UpdateDateTimeVal(newestDateTimeString, ourDateTimeNode);
+			UpdateDateTimeVal(newestDateTimeString, theirDateTimeNode);
 		}
 
-		private static void RestoreOriginalIfTimestampIsTheOnlyChange(XmlNode ancestorDateTimeNode, XmlNode otherDateTimeNode)
+		private static bool RestoreOriginalIfTimestampIsTheOnlyChange(XmlNode ancestorDateTimeNode, XmlNode otherDateTimeNode)
 		{
-			if (ancestorDateTimeNode == null || otherDateTimeNode == null)
-				return;
+			if (ancestorDateTimeNode == null)
+				return false;
+
+			if (otherDateTimeNode == null)
+				return true;
 
 			// Values that are are the same are not of interest.
 			var ancestorAttr = ancestorDateTimeNode.Attributes["val"];
 			var otherAttr = otherDateTimeNode.Attributes["val"];
 			if (ancestorAttr.Value == otherAttr.Value)
-				return;
+				return true;
 
 			// Get parents of both nodes
 			var ancestorDateTimeNodeParent = ancestorDateTimeNode.ParentNode;
@@ -54,10 +70,11 @@ namespace LibFLExBridgeChorusPlugin.Handling
 			otherAttr.Value = ancestorAttr.Value;
 
 			if (XmlUtilities.AreXmlElementsEqual(ancestorDateTimeNodeParent, otherDateTimeNodeParent))
-				return; // Only change was the timestamp, so keep it.
+				return true; // Only change was the timestamp, so keep it.
 
 			// Restore the original value.
 			otherAttr.Value = originalOtherValue;
+			return false;
 		}
 
 		private static void UpdateDateTimeVal(string newest, XmlNode currentDateTimeNode)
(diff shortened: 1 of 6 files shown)


ermshiperete:
Tested S/R with a locally build version and went through the failing scenarios of the [test plan](https://docs.google.com/a/sil.org/spreadsheets/d/1SNwVnW9_r9-365uOHMnc_Fr7EU4DECdERupMq88SkTs/edit?usp=sharing). The scenarios "Edit conflicts: FLEx edits entry, LF edits a different entry field" and "Edit conflicts: FLEx edits sense, LF edits a different meaning field" work now.

"Edit conflicts: FLEx edits sense, LF deletes meaning" still fails.

ermshiperete:
With the latest update of this PR also the "Edit conflicts: FLEx edits sense, LF deletes meaning" scenario
works now.

jasonleenaylor:
This feature with the date was very intentional. It was designed to avoid conflicts and reduce merges related to the DateModified. If the DateModified is changed after the merge it would increase the number of merges that have to happen in order to stabilize a change.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-:-Ka2duDKX1LByPmTJAww:bromgvi)*

jasonleenaylor:
I'll need to look at this carefully to make sure that it doesn't bring back the problems which were solved by previous date code.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-:-Ka2eyOefkBBETbthYZp:b-3kdssp)*

ermshiperete:
I'm aware that the date feature was intentionally and I tried carefully to change it as little as possible.

I guess it all depends on the purpose of the DateModified field. So far I assumed that if I see a timestamp in DateModified and it hasn't changed than I can assume that the other fields in the entry didn't change as well. This assumption proved to be not true with the implementation we had so far: it was true most of the time, but if the remote and local changes resulted in a non-conflicting merge, some fields in the entry changed but the DateModified didn't.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-:-Ka7qKv3mTzx1G7CVK8j:b466i1j)*

regnrand:
The 'easiest thing' would be to simply update the DateModified property to the current UTC time, when doing a merge.

1. DateModified is to be updated when something changes at the root node or any lower contained node.
2. Normally, that would be by some human editor, but the merger code is an 'editor', if not a human one.

So, feel free to dump all that complex new diff code and just plop in the current UTC time. You only get to this code in a merge. I'm not even sure the calls to "RestoreOriginalIfTimestampIsTheOnlyChange" are worth the bother, although LF may end up with some no-op work by not restoring them, if both systems had only updated the DateModified property.

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-:-Ka84FjIOSw2ZBVX4ssB:b-mdjemm)*

megahirt:
I know that we discussed this functionality several weeks ago, and I really appreciate your effort Eberhard in implementing what you have.  If I recall correctly, LF assumes that merges are "edits" to the model and that the date modified will be updated to reflect that edit.

I remember thinking in those office conversations exactly what Randy has stated, that the easiest solution is to just update DateModified to UTC now().  This makes the code much simpler, but it assumes that this would only be executed in a merge conflict situation.  I believe that's the context of this change.

If RestoreOriginalIfTimestampIsTheOnlyChange is expensive and adds complexity, I also agree that it's not worth doing.  I think it is the *right* thing to do, but seems like a rare case for an attribute (DateModified) that the user doesn't care about.  If there's a cheap way to answer that question, then go with that.  Otherwise, I'm in favor of a two line solution:
UpdateDateTimeVal(DateTime.UtcNow, ourDateTimeNode);
UpdateDateTimeVal(DateTime.UtcNow, theirDateTimeNode);

Again Eberhard, I appreciate the time you've put into this PR and functionality.  This is a core change and testing will need to be done on both our side (Language Forge S/R) and with FLEx to verify that this is an improvement and doesn't re-introduce former merge issues.

---

Reviewed 1 of 5 files at r1, 5 of 5 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-:-KaofiuP9S0EfpT_J9E3:bjx2mtr)*

regnrand:
One other option would be to let the original code do its thing and set DateModified to the newest between ours and theirs. Then, do this comparison on those two nodes using something like:

if (!XmlUtilities.AreXmlElementsEqual(ourDateTimeNodeParent, theirDateTimeNodeParent))
{
    // Give up and set DateModified to DateTime.UtcNow on both, since we updated both to the newest,
    // but there are still other differences, and LF needs to know if it has stale data.
}

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-:-KapC97g0hN1TNJJXRev:b-c6yo4p)*

megahirt:
@jasonleenaylor would you be able to look at this PR again and give your feedback?  I'm sorry I am digging this one up after not having reviewed it for several weeks.  thanks for your help.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-:-Kapk3ZkS7US9UEkt2C6:b-yg8h43)*

ermshiperete:
Simplified the code by updating DateModified always in case of a merge (unless the only change is a change in DateModified). Unfortunately this required changes to even more tests...

The original implementation of this PR is still available as r2 in reviewable.io.

@jasonleenaylor, @megahirt, can you please take another look at the new version of this PR? Thanks!

---

Review status: 0 of 6 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-:-KbUVI3lM9tR7zk68zFH:b-d527bn)*

cambell-prince:
Reviewed 7 of 7 files at r3.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[src/LibFLExBridge-ChorusPlugin/Handling/PreferMostRecentTimePreMerger.cs, line 49 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-KbweMTrmUXqlZGTPA97:-KbweMTrmUXqlZGTPA98:b-s8o6p0) ([raw file](https://github.com/sillsdev/flexbridge/blob/ee19cfc0b9e3f4e58c01473a10de4d4277f356c4/src/LibFLExBridge-ChorusPlugin/Handling/PreferMostRecentTimePreMerger.cs#L49)):*
> ```Smalltalk
> 			UpdateDateTimeVal(newestDateTimeString, theirDateTimeNode);
> 		}
> ```

This approach looks good to me.  If data has changed as a result of the merge setting date modified to now seems more appropriate than choosing either ours or theirs.

---

*[src/LibFLExBridge-ChorusPluginTests/Handling/BaseFieldWorksTypeHandlerTests.cs, line 59 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-KbwfJNVtHXZn1WT9YHw:-KbwfJNVtHXZn1WT9YHx:boms4c) ([raw file](https://github.com/sillsdev/flexbridge/blob/ee19cfc0b9e3f4e58c01473a10de4d4277f356c4/src/LibFLExBridge-ChorusPluginTests/Handling/BaseFieldWorksTypeHandlerTests.cs#L59)):*
> ```Smalltalk
> 			}
> 		}
> 	}
> ```

Don't we have a date time extension for this?  No objection to it being here though.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137)*

megahirt:
:lgtm:

---

Reviewed 6 of 7 files at r3.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-:-Kbwl45Nl6-YFn5V1LSx:bnfp4nl)*

jasonleenaylor:
I didn't see anything to hold this back.

ermshiperete:
Review status: all files reviewed at latest revision, 1 unresolved discussion, all commit checks successful.

---

*[src/LibFLExBridge-ChorusPluginTests/Handling/BaseFieldWorksTypeHandlerTests.cs, line 59 at r3](https://reviewable.io:443/reviews/sillsdev/flexbridge/137#-KbwfJNVtHXZn1WT9YHw:-KbzOeIaadfSQkLvZ1lT:bodhnl4) ([raw file](https://github.com/sillsdev/flexbridge/blob/ee19cfc0b9e3f4e58c01473a10de4d4277f356c4/src/LibFLExBridge-ChorusPluginTests/Handling/BaseFieldWorksTypeHandlerTests.cs#L59)):*
<details><summary><i>Previously, cambell-prince (Cambell) wrote…</i></summary><blockquote>

Don't we have a date time extension for this?  No objection to it being here though.
</blockquote></details>

Unfortunately not. The extension method uses the `yyyy-MM-ddTHH:mm:ssZ` format which is slightly different to what Flexbridge writes (`yyyy-M-d H:m:s.FFF`).

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/137)*

PR #140, opened 2017-02-17
SendReceive User CSS Overrides

papeh:
* Include ConfigurationSettings/**Overrides.css in Send/Receive
* Alphabetize excluded extensions

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/140)


jasonleenaylor:
I'm not sure the quality of merge we will get with the .css files. We may want to do all or nothing and I think that requires a handler right?

jasonleenaylor:
After further research and discussion with @regnrand we've decided that the default behavior here (which will generate a conflict but keep the S/R initiators file) is probably sufficient. LGTM

PR #149, opened 2017-06-21
Send/Receive Chorus Notes with LF comments

rmunn:
This PR needs to be merged before the corresponding LfMerge PR is merged, since LfMerge will depend on having an LfMergeBridge DLL that knows the new GetChorusNotes and WriteToChorusNotes actions.

There are no unit tests yet, but I'm creating the PR anyway so that we can kick off a TeamCity build and make sure that it compiles.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/149)


rmunn:
In the interests of getting this merged into master so that we can test it manually on the dev server, I'm going to merge this even though there are no unit tests yet.

PR #169, opened 2017-10-09
Remove obsolete checked-in binaries

ermshiperete:
We replaced the DNZip task with the Unzip task of the
MSBuild.Community.Tasks and we no longer use the
MSBuild.ExtensionPack, so we don't need these binaries anymore. If
we ever need them again we should get them as nuget packages instead
of checking them into the repo.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/169)


papeh:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/169#-:-Kw0l9iq3dojdPcuX63D:bnfp4nl)*

PR #174, opened 2017-11-16
Use the NUnitTeamCity task when running on TC

jasonleenaylor:
* Starting on Oct 10 anything output stderr failed a build
  This avoids 'plugin loaded...' messages that were being
  output to stderr in the build.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/174)


jasonleenaylor:
TC Build is busted and I can't test this without merging it really. Doesn't break testing on a developer machine.

PR #178, opened 2017-12-05
Make Installer Work

papeh:
* Copy Mercurial to the directory harvested by the installer
* Exclude test code and the RepositoryUtility from this copy
* Write to the registry key FLEx expects

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/178)


papeh:
@jasonleenaylor, the LocalAppData environment var on the user's machine is part of the path to where we store settings. However, since we don't need to specify this in the installer, I hope to disable this portion of the installer prior to Beta testing (at the very least before stable).

PR #186, opened 2018-01-10
Fix fixutf8

marksvc:
The fixutf8 extension appears to start off enabled with a relative
path. This doesn't work on Linux, and just adding a second line with
an absolute path doesn't fix it. So comment out any existing fixutf8
extension lines so we only have the one line with the absolute path.
Fixes LT-18796.

### DIFF
--- a/debian/postinst
+++ b/debian/postinst
@@ -25,6 +25,8 @@ case "$1" in
 			ARCH="$(/usr/bin/arch)"
 			unzip Mercurial-${ARCH}.zip
 			chmod +x Mercurial/hg
+			# Turn off any existing fixutf8 extensions with potentially problematic paths
+			perl -pi -e "/^fixutf8/ && s/^/#/" Mercurial/mercurial.ini
 			echo "fixutf8=/usr/lib/flexbridge/MercurialExtensions/fixutf8/fixutf8.py" >> Mercurial/mercurial.ini
 		fi
 


jasonleenaylor:
It might be better to remove the problematic line from the Mercurial.zip in Chorus.

---







*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/186#-:-L2WyDvb4RC_cN6YbeNt:byajbf)*

ermshiperete:
I agree with Jason. However, as additional safety I like having that line in the `postinst` script - it's too easy that that line slips in again.
Another option would be to create a new `mercurial.ini` in `rules` or `Makefile` (cf [LfMerge](https://github.com/sillsdev/LfMerge/blob/master/debian/rules.in#L18)).

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/186#-:-L2Z28YTMXU1vMKTsjx6:b2pvcm0)*

marksvc:
Based on various discussion, I am removing the line using postinst, and plan to remove the line from the source.

ermshiperete:
:lgtm: (I guess you checked that Perl is installed on a fresh Ubuntu installation? Otherwise we'd have to add it as a dependency to the `control` file.)

---

Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/186#-:-L2dLfvbrXv7a_TrmsYf:bshw0p9)*

PR #187, opened 2018-01-17
Fix LT-18788: Not able to use FLEx Bridge 3.0 If we try to export Lex…

JamesPrabu:
* Given meaningful message and fixed crash

### DIFF
--- a/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs
+++ b/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs
@@ -33,7 +33,8 @@ namespace FLEx_ChorusPlugin.Infrastructure.ActionHandlers
 		{
 			if (!UpdateBranchHelper.UpdateToTheCorrectBranchHeadIfPossible(new FlexUpdateBranchHelperStrategy(), commandLineArgs["-fwmodel"], cloneResult, cloneLocation))
 			{
-					cloneResult.Message = CommonResources.kFlexUpdateRequired;
+                if (string.IsNullOrEmpty(cloneResult.Message))
+                    cloneResult.Message = CommonResources.kFlexUpdateRequired;
 			}
 		}
 
@@ -75,7 +76,12 @@ namespace FLEx_ChorusPlugin.Infrastructure.ActionHandlers
 					Directory.Delete(cloneLocation, true);
 					_newFwProjectPathname = null;
 					return;
-				case FinalCloneResult.Cloned:
+                case FinalCloneResult.ChosenRepositoryIsEmpty:
+                    MessageBox.Show(actualCloneResult.Message, CommonResources.kObtainProject, MessageBoxButtons.OK, MessageBoxIcon.Exclamation);
+                    Directory.Delete(cloneLocation, true);
+                    _newFwProjectPathname = null;
+                    return;
+                case FinalCloneResult.Cloned:
 					_gotClone = true;
 					break;
 			}
--- a/src/LibTriboroughBridge-ChorusPlugin/Infrastructure/FinalCloneResult.cs
+++ b/src/LibTriboroughBridge-ChorusPlugin/Infrastructure/FinalCloneResult.cs
@@ -7,6 +7,7 @@ namespace LibTriboroughBridgeChorusPlugin.Infrastructure
 	{
 		Cloned,
 		ExistingCloneTargetFolder,
-		FlexVersionIsTooOld
-	}
+		FlexVersionIsTooOld,
+        ChosenRepositoryIsEmpty
+    }
 }
--- a/src/LibTriboroughBridge-ChorusPlugin/Infrastructure/UpdateBranchHelper.cs
+++ b/src/LibTriboroughBridge-ChorusPlugin/Infrastructure/UpdateBranchHelper.cs
@@ -87,13 +87,18 @@ namespace LibTriboroughBridgeChorusPlugin.Infrastructure
 					allHeads.Remove(LibTriboroughBridgeSharedConstants.Default);
 				}
 
-				// 'default' is no longer present in 'allHeads'.
-				// If all of them are higher, then it is a no go.
-				if (allHeads.Count == 0)
-				{
-					// No useable model version, so bailout with a message to the user telling them they are 'toast'.
-					updateResult.FinalCloneResult = FinalCloneResult.FlexVersionIsTooOld;
-					Directory.Delete(repoPath, true);
+                // 'default' is no longer present in 'allHeads'.
+                // If all of them are higher, then it is a no go.
+                if (allHeads.Count == 0)
+                {
+                    updateResult.FinalCloneResult = FinalCloneResult.ChosenRepositoryIsEmpty;
+                    updateResult.Message = "The chosen repository is empty. Send/Receive for the first time has not yet been used.";
+                    return false;
+                }
+                else
+                {
+                    // No useable model version, so bailout with a message to the user telling them they are 'toast'.
+                    updateResult.FinalCloneResult = FinalCloneResult.FlexVersionIsTooOld;
 					return false;
 				}
 


jasonleenaylor:
:lgtm:

---

Reviewed 3 of 3 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/187#-:-L39GDtkX04nQsH5-RUd:bnfp4nl)*

marksvc:
Review status: all files reviewed at latest revision, 4 unresolved discussions, all commit checks successful.

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs, line 36 at r1](https://reviewable.io:443/reviews/sillsdev/flexbridge/187#-L39G_9SxdU5bpIg44Pb:-L39G_9SxdU5bpIg44Pc:b42jhci) ([raw file](https://github.com/sillsdev/flexbridge/blob/7238e43d9f1a3e9a82fc1f941604827d6af29c66/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs#L36)):*
> ```Smalltalk
> 			if (!UpdateBranchHelper.UpdateToTheCorrectBranchHeadIfPossible(new FlexUpdateBranchHelperStrategy(), commandLineArgs["-fwmodel"], cloneResult, cloneLocation))
> 			{
>                 if (string.IsNullOrEmpty(cloneResult.Message))
> ```

Why is this check done? Is it so an existing message isn't overwritten? It seems like having a check here might prevent us from seeing the kFlexUpdateRequired message sometimes, but is that not a problem do you think?

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs, line 79 at r1](https://reviewable.io:443/reviews/sillsdev/flexbridge/187#-L39Hvjvkeqmg-wAQEPh:-L39Hvjvkeqmg-wAQEPi:bkl2j9b) ([raw file](https://github.com/sillsdev/flexbridge/blob/7238e43d9f1a3e9a82fc1f941604827d6af29c66/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs#L79)):*
> ```Smalltalk
> 					_newFwProjectPathname = null;
> 					return;
>                 case FinalCloneResult.ChosenRepositoryIsEmpty:
> ```

There is a C# language feature you can use if you want FinalCloneResult.FlexVersionIsTooOld and FinalCloneResult.ChosenRepositoryIsEmpty to do the same thing. You can write code this way, stacking case statements:

    switch (foo)
    {
        case A:
            blah;
            return;
        case B:
        case C:
            return;
    }

So your code could be simplified like so, by combining/stacking the two case statements:

    ...
    case FinalCloneResult.FlexVersionIsTooOld:
    case FinalCloneResult.ChosenRepositoryIsEmpty:
        MessageBox.Show(actualCloneResult.Message,
            CommonResources.kObtainProject, MessageBoxButtons.OK,
            MessageBoxIcon.Exclamation);
        Directory.Delete(cloneLocation, true);
        _newFwProjectPathname = null;
    ...

---

*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs, line 83 at r1](https://reviewable.io:443/reviews/sillsdev/flexbridge/187#-L39GrTCkSTzT4u8QOGx:-L39GrTCkSTzT4u8QOGy:b-al43m3) ([raw file](https://github.com/sillsdev/flexbridge/blob/7238e43d9f1a3e9a82fc1f941604827d6af29c66/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs#L83)):*
> ```Smalltalk
>                     Directory.Delete(cloneLocation, true);
>                     _newFwProjectPathname = null;
>                     return;
> ```

It looks like some tabs have been replaced by spaces. I wonder if your IDE needs some setting made to be using tabs in these C# files?

---

*[src/LibTriboroughBridge-ChorusPlugin/Infrastructure/UpdateBranchHelper.cs, line 95 at r1](https://reviewable.io:443/reviews/sillsdev/flexbridge/187#-L39K3sjC7A4wY7NUYb8:-L39K3sjC7A4wY7NUYb9:b-j4lzzh) ([raw file](https://github.com/sillsdev/flexbridge/blob/7238e43d9f1a3e9a82fc1f941604827d6af29c66/src/LibTriboroughBridge-ChorusPlugin/Infrastructure/UpdateBranchHelper.cs#L95)):*
> ```Smalltalk
>                 {
>                     updateResult.FinalCloneResult = FinalCloneResult.ChosenRepositoryIsEmpty;
>                     updateResult.Message = "The chosen repository is empty. Send/Receive for the first time has not yet been used.";
> ```

Are you sure this is what triggers this condition? It looks like the intent of the code in this area was to detect if there is no head with a version that flexbridge can work with, and if so to tell the user that their flex/flexbridge is too old to work with the data available. 

From your error message above, it sounds like you have found that this part of the code can also be reached in the event that there is no data in the repository at all. If so, I think you'll want to differentiate between the "there is no data at all" condition and the "there is data but it's for version of flex that you aren't able to handle" condition. 

Also, as it is, the if/else in this patch will never let the rest of this method run, even in non-problem conditions.

(And I'm not sure the Directory.Delete line should have been deleted; was that causing problems?)

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/187)*

jasonleenaylor:
*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs, line 79 at r1](https://reviewable.io:443/reviews/sillsdev/flexbridge/187#-L39Hvjvkeqmg-wAQEPh:-L39pEv3SGU4j_gflVGg:b-yj00jg) ([raw file](https://github.com/sillsdev/flexbridge/blob/7238e43d9f1a3e9a82fc1f941604827d6af29c66/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs#L79)):*
<details><summary><i>Previously, marksvc wrote…</i></summary><blockquote>

There is a C# language feature you can use if you want FinalCloneResult.FlexVersionIsTooOld and FinalCloneResult.ChosenRepositoryIsEmpty to do the same thing. You can write code this way, stacking case statements:

    switch (foo)
    {
        case A:
            blah;
            return;
        case B:
        case C:
            return;
    }

So your code could be simplified like so, by combining/stacking the two case statements:

    ...
    case FinalCloneResult.FlexVersionIsTooOld:
    case FinalCloneResult.ChosenRepositoryIsEmpty:
        MessageBox.Show(actualCloneResult.Message,
            CommonResources.kObtainProject, MessageBoxButtons.OK,
            MessageBoxIcon.Exclamation);
        Directory.Delete(cloneLocation, true);
        _newFwProjectPathname = null;
    ...
</blockquote></details>

Actually Mark, the new case is purposefully skipping the Directory.Delete, not clear from the diff.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/187)*

jasonleenaylor:
*[src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs, line 79 at r1](https://reviewable.io:443/reviews/sillsdev/flexbridge/187#-L39Hvjvkeqmg-wAQEPh:-L39pgdrZAAXqacrnx1W:bgqu84c) ([raw file](https://github.com/sillsdev/flexbridge/blob/7238e43d9f1a3e9a82fc1f941604827d6af29c66/src/FLEx-ChorusPlugin/Infrastructure/ActionHandlers/ObtainProjectStrategyFlex.cs#L79)):*

Actually Mark, the new case is purposefully skipping the Directory.Delete, not clear from the diff.
</blockquote></details>

Hmm, no, it is the same. I think it was supposed to be different, disregard.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/187)*

PR #190, opened 2018-01-18
Ignore problematic DllImport references

marksvc:
MAPI32.DLL was Windows-only in MAPIHelper.cs before #if !MONO was
removed.
__Internal probably should have been ignored by dh_clideps without us
having to say so. This is in Platform.cs.
This fixes a Linux packaging error.

Both Platform.cs and MAPIHelper.cs are in libpalaso, and this problem
was introduced to flexbridge packaging when we switched to use the
master branch of libpalaso.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/190)


papeh:
Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/190)*

PR #193, opened 2018-01-29
Fix geckofx crash on development machine

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/193)

### DIFF
--- a/flexbridge
+++ b/flexbridge
@@ -4,6 +4,11 @@
 
 set -e -o pipefail
 
+# On a developer machine, MONO_ENVIRON may be set, and to the location of the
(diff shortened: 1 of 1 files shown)


papeh:
That's certainly interesting. Thanks for figuring that out. I can't think of any cases where this would  break anything, since we're not exporting the `unset`, so :lgtm:

---

Reviewed 1 of 1 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/193#-:-L43lR1411YcFTRFvkkD:b9e9et5)*

PR #196, opened 2018-02-05
Compile using geckofx45 from nuget

marksvc:
Use nuget to get geckofx45 to compile, rather than using it from
FieldWorks since we don't necessarily know where that will be on a
developer's machine. This will also be used at run-time on a
developer's machine.

But don't ship geckofx45, since flexbridge is an architecture-
independent package. Instead, use geckofx45 from FieldWorks at run-
time, when flexbridge is installed from a package.

* Use nuget to download geckofx45.
* Copy Geckofx-Core.dll.config to where its assembly is.
* Set environment variable to find and use the right gecko .dll and
.so files at compile-time and run-time.
* Leave environ and environ-xulrunner in the root of the VCS, and use
them from there when running on a development machine, to ease
specifying paths in environ-xulrunner.
* Give hintpaths in csproj files. Just referencing 64-bit assembly, as
we don't ship the geckofx assemblies anyway. (Also, we only build
packages in 64-bit environments.)
* Set csproj files to not copy geckofx assemblies into flexbridge
output folder, by setting Private to False.
* Remove geckofx reference from FLEx-ChorusPlugin, as it appears to be
unused.
* Remove geckofx29 package build dependency.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/196)


RaymondLuong3:
:lgtm: But I'm sure someone else might want to take a look too

---

Review status: 0 of 9 files reviewed at latest revision, 1 unresolved discussion.

---

*[src/FLEx-ChorusPlugin/FLEx-ChorusPlugin.csproj, line 163 at r1](https://reviewable.io:443/reviews/sillsdev/flexbridge/196#-L4bvl55G3vOsqHZSFiZ:-L4bvl55G3vOsqHZSFi_:b-e1me5h) ([raw file](https://github.com/sillsdev/flexbridge/blob/267b0c33ff0884a698d0d0b847fc6ebf377d5091/src/FLEx-ChorusPlugin/FLEx-ChorusPlugin.csproj#L163)):*
> ```XML
>       <SpecificVersion>False</SpecificVersion>
>       <Package>Geckofx-Winforms</Package>
>     </Reference>
> ```

Good catch, no need for it if it's not being used.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/196#-:-L4bw9Uz7WpiFW-H_5mC:bdqf3eo)*

ermshiperete:
You're changes :lgtm:, but I noticed two things that you might want to change to clean-up `run-app` while we're touching that file. However, that's up to you :smile: 

---

Reviewed 6 of 9 files at r1, 5 of 5 files at r2.

---

*[lib/common/run-app, line 43 at r2](https://reviewable.io:443/reviews/sillsdev/flexbridge/196#-L4eG_NWwZjOWm_Ng8CF:-L4eG_NWwZjOWm_Ng8CG:b4itliq) ([raw file](https://github.com/sillsdev/flexbridge/blob/567be1d5bd4b1706dc412fb3b4a060779354efbe/lib/common/run-app#L43)):*
> MONO_PREFIX=/usr/local

This seems wrong. Does anyone still have mono in `/usr/local/`? And we already referred to MONO_PREFIX at the top of the file. I think this line should go, and be replaced with a line at the top of the file setting MONO_PREFIX to /opt/mono-sil (or whatever mono version we're expecting), but only if MONO_PREFIX isn't set already.

---

*[lib/common/run-app, line 52 at r2](https://reviewable.io:443/reviews/sillsdev/flexbridge/196#-L4eH-FhHummqUlc8MCf:-L4eH-FhHummqUlc8MCg:b97q7dr) ([raw file](https://github.com/sillsdev/flexbridge/blob/567be1d5bd4b1706dc412fb3b4a060779354efbe/lib/common/run-app#L52)):*
> PKG_CONFIG_PATH

We have the same line three lines above, so I'd suggest to delete this one.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/196#-:-L4eFVuZPoTGIM498ogB:bv85qtc)*

marksvc:
Thanks guys. I'll plan to adjust run-app in a followup commit.

---

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/196#-:-L4gU38Kc9gMy18bKq-o:bybqnj)*

PR #200, opened 2018-02-09
Fix LT-18918:Update FlexBridge to support new DataModel

vkarthim:
*Added code to remove ReversalEntries property in LexSense
*Added code to add Senses property in ReversalIndexEntry
*Updated the ModelVersion number
*Added the test case for this case

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/200)


src/LibFLExBridge-ChorusPluginTests/Integration/UpdateMetaDataCacheTests.cs
```diff
@@ -333,6 +333,16 @@ public void MetaDataCacheIsUpdated()
 
 			// 7000071: No actual model change.
 			CheckNoModelChangesUpgrade(mdc, fileHandler, 7000071);
+
+			// 7000072:
+			// Make sure prior model is expected.
+			//Removing ReversalEntries property from LexSense from owning collection to owning sequence.
```
jasonleenaylor: Remove part of the comment leaving: //Removing ReversalEntries property from LexSense

PR #211, opened 2018-03-05
Bump version to 3.0.1

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/211)


src/Installer/Stable.htm
```diff
@@ -1,6 +1,15 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <html>
   <div class="releasenotes">
+    <h2>3.0.1 05-Mar-2018</h2>
```
jasonleenaylor: Not that this matters much, but it should probably be the Beta, or the Alpha .htm which was targeted by the script here.

marksvc: I will re do that, then. Should it be Beta or Alpha?

jasonleenaylor: Lets use Beta.

ermshiperete:
Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/sillsdev/flexbridge/211)*

PR #220, opened 2018-03-28
Fix bug when annotations contained no messages

rmunn:
Some annotations in real user data have NO `<message>` elements inside them, which causes the merge to fail and the project to be put on hold. This is not *supposed* to happen, but it does, so we need to deal with it. This commit is the simplest way I've found to do so: it skips any such annotations, because we can't extract any meaningful data from them anyway (and trying to do so ends up causing other issues).

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/220)


ermshiperete:
:lgtm:

---

Reviewed 2 of 2 files at r1.

---



*Comments from [Reviewable](https://beta.reviewable.io/reviews/sillsdev/flexbridge/220#-:-L8foWDN9qfyQlan_1Ik:bnfp4nl)*

PR #237, opened 2018-08-07
Fix: LT-19303 FLEx Bridge should install under Progra

samdoss:
* Need to set Default AppFolder Directory when it run quite mode

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/237)


samdoss:
Yes Hasso, Its required when it is doing not in quietmode.

PR #248, opened 2019-07-17
Fix build

ermshiperete:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/248)


rmunn:
For future reference, the reason why we switched to GetVersionTask.alt was https://github.com/GitTools/GitVersion/issues/1483, which is supposedly fixed in GitVersion 5.0.0. (I have not yet verified this). So once we verify that switching the build to GitVersion 5 fixes the issue, we might be able to switch back to the main GitVersion package and no longer have to maintain a fork.

PR #271, opened 2020-04-15
Fix L10nSharp load problem

papeh:
Download L10NSharp.dll from TeamCity Palaso build.
Because everyone (at least Chorus and FLEx) uses L10NSharp.dll
from Palaso builds, there will be no version conflicts.
(This is what FLEx Bridge builds were already doing on TeamCity.)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/271)


ermshiperete:
Wouldn't it be better to change libpalaso and chorus to use the L10NSharp nuget package? That way you could compare the L10NSharp package version used in all three projects. Currently you'll have to check the TC dependencies on all three plus the TC L10NSharp build, and then this still doesn't tell you if what you have on your machine matches matches that version. Things get even more interesting when multiple branches come into play.

PR #263, opened 2020-03-23
Add to FlexBridge a version number (LT-20019)

JohnThomson:
The branch name that FlexBridge uses will now be made up of ModelVersion.FlexBridgeDataVersion

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/263)

PR #283, opened 2020-06-11
Fix model version parsing after #263

rmunn:
Now that model versions are doubles, we need to change some of the int.Parse calls in LfMergeBridge.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/283)

### DIFF
--- a/src/LfMergeBridge/LanguageForgeMakeCloneActionHandler.cs
+++ b/src/LfMergeBridge/LanguageForgeMakeCloneActionHandler.cs
@@ -109,7 +109,7 @@ namespace LfMergeBridge
 			}
 			// See if repo has higher branch than LF called for.
 			var highestHead = LfMergeBridgeUtilities.GetHighestRevision(hgRepository);
-			if (int.Parse(highestHead.Branch) > int.Parse(desiredBranchName))
+			if (double.Parse(highestHead.Branch) > double.Parse(desiredBranchName))
 			{
 				// Clone has a higher data model than LF asked for.
 				Directory.Delete(actualClonePath, true);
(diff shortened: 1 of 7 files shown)


src/LfMergeBridge/LanguageForgeMakeCloneActionHandler.cs
```diff
@@ -109,7 +109,7 @@ private static void FinishClone(IProgress progress, ref string somethingForClien
 			}
 			// See if repo has higher branch than LF called for.
 			var highestHead = LfMergeBridgeUtilities.GetHighestRevision(hgRepository);
-			if (int.Parse(highestHead.Branch) > int.Parse(desiredBranchName))
+			if (double.Parse(highestHead.Branch) > double.Parse(desiredBranchName))
```
ermshiperete, line 112: Probably use `GetModelVersionFromBranchName` here as well?

jasonleenaylor:
I'm confused about this PR. Why is lfmerge so far behind develop? Aren't all the necessary changes already on the develop branch?

rmunn:
Back when I last looked at the FLExBridge code, we were building the LfMergeBridge code on a separate `lfmerge` branch. I saw that it had been deleted from the FLExBridge GitHub repo, so I restored it before adding this patch, because I assumed that the `master` and `develop` branches did not contain the LfMergeBridge code. But if it was merged into `develop` at some point during the year I was on furlough and not paying attention to the repo, then I'll do a build of LfMergeBridge from the `develop` branch and see if that solves the problem.

rmunn:
Closing; will open a new PR against `develop` branch with the actual changes that need to happen.

rmunn:
Replaced and superseded by #285.

PR #285, opened 2020-06-12
Fix model version parsing after #263

rmunn:
Now that model versions are doubles, we need to change some of the int.Parse calls in LfMergeBridge.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/285)

### DIFF
--- a/src/LfMergeBridge/LanguageForgeMakeCloneActionHandler.cs
+++ b/src/LfMergeBridge/LanguageForgeMakeCloneActionHandler.cs
@@ -180,7 +180,8 @@ namespace LfMergeBridge
 
 			var user = options.ContainsKey(LfMergeBridgeUtilities.user) ? options[LfMergeBridgeUtilities.user] : null;
 
-			FinishClone(progress, ref somethingForClient, cloneBase, actualClonePath, options[LfMergeBridgeUtilities.fdoDataModelVersion], user, DeleteRepoIfNoSuchBranch(options));
+			var desiredBranchName = FlexBridgeConstants.FlexBridgeDataVersion + "." + options[LfMergeBridgeUtilities.fdoDataModelVersion];
+			FinishClone(progress, ref somethingForClient, cloneBase, actualClonePath, desiredBranchName, user, DeleteRepoIfNoSuchBranch(options));
 		}
 
 		/// <summary>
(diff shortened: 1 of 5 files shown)


src/LfMergeBridge/LanguageForgeMakeCloneActionHandler.cs
```diff
@@ -178,7 +180,8 @@ void IBridgeActionTypeHandler.StartWorking(IProgress progress, Dictionary<string
 
 			var user = options.ContainsKey(LfMergeBridgeUtilities.user) ? options[LfMergeBridgeUtilities.user] : null;
 
-			FinishClone(progress, ref somethingForClient, cloneBase, actualClonePath, options[LfMergeBridgeUtilities.fdoDataModelVersion], user, DeleteRepoIfNoSuchBranch(options));
+			var desiredBranchName = FlexBridgeConstants.FlexBridgeDataVersion + "." + options[LfMergeBridgeUtilities.fdoDataModelVersion];
```
ermshiperete, line 183: It seems to me this (and similar places where we calculate the branch name) should be refactored so that there is a method that takes a model version and returns the branch name, kind of the opposite of  `GetModelVersionFromBranchName`. This could maybe be added to `IUpdateBranchHelperStrategy`.

rmunn:
@ermshiperete - I'd like you to take another look at this, especially commit 92aca79. I'm not 100% confident that what I did in [LanguageForgeMakeCloneActionHandler.cs](https://github.com/sillsdev/flexbridge/commit/92aca7939d5b972651ae6f89dd0703f2265b24d2#diff-8e6b40d7be5f97b4e008975d231591fbR89) is the right approach to handling "older" repos such as the `testlangproj` repo in the LfMerge unit tests. Yes, it gets the LfMerge unit tests to pass (once https://github.com/sillsdev/LfMerge/pull/100 is applied), but is it the right solution for other situations? I'm trying to do tests with FieldWorks Send/Receive, but I'd much rather have more eyes on this than just mine.

rmunn:
@ermshiperete Does this still LGTY after commit https://github.com/sillsdev/flexbridge/commit/e3270097e75ce88ce887fb835f7b26c0bb399ab7?

PR #311, opened 2021-04-20
Fix Crashes on Linux

papeh:
Use the environment variables provided by FieldWorks
without modification.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/311)

### DIFF
--- a/debian/control
+++ b/debian/control
@@ -1,7 +1,7 @@
 Source: flexbridge
 Section: misc
 Priority: extra
-Maintainer: Stephen McConnel <stephen_mcconnel@sil.org>
+Maintainer: FLEx Dev Team <FLEx_DevTeam@sil.org>
 Build-Depends: debhelper (>= 9.0.0),
  cli-common-dev (>= 0.8~),
  mono5-sil,
@@ -30,7 +30,7 @@ Depends: ${misc:Depends}, ${cli:Depends},
  unzip,
  adduser (>= 3.11)
 Description: Allow multiple FieldWorks users to collaborate remotely
- FLEx Bridge is an application that allows multiple FieldWorks 8.0 users
- to do remote collaboration (i.e., not necessarily connected by a local
+ FLEx Bridge is an application that allows multiple FieldWorks >=8.0 users
+ to collaborate remotely (i.e., not necessarily connected by a local
  network).  Data can be shared via a local network (which is not really
  "remote"), the Internet, or a USB memory device.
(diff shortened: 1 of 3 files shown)


marksvc (commented):
:eyes: 

> Reviewed 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/311#-:-MYjmXuO697LDddL_tEf:bd5ty59)* status: 1 of 3 files reviewed, all discussions resolved (waiting on @papeh)

marksvc (commented):
I looked over the flexbridge environ file to see what we might be missing that we don't want to. A couple things stood out to me that aren't also in the fieldworks environ file. 

The first is that the location of the flexbridge exe is put into LD_LIBRARY_PATH, MONO_PATH, and PATH. LD_LIBRARY_PATH will only be significant if there are .so files in the location*. MONO_PATH will only(?) be significant if there are dll files in the location that otherwise fail to be loaded. I'm not sure that PATH would be significant. So if anything starts to go funny with not finding a library, it may be due to this.

*Flexbridge probably was providing a Firefox-Linux64 directory of .so files. That won't be significant now since the fieldworks ones are the ones that will be used. This also means that we can stop shipping the gecko .so files with Linux flexbridge.

```bash
# Add both the user installation directory and the developer build output path to the paths
# (so things work in both situations).
LD_LIBRARY_PATH="${BASE}:${BASE}/output/${BUILD}:${LD_LIBRARY_PATH}"
MONO_PATH="${BASE}:${BASE}/output/${BUILD}:${MONO_PATH}"
PATH="${BASE}/output/${BUILD}:${PATH}"
```

The second item in the flexbridge environ file is

```bash
# set HGRCPATH so that we ignore ~/.hgrc files which might have content that is
# incompatible with our version of Mercurial
export HGRCPATH=
```

It looks like we will want to keep this. It could be put into the fieldworks environ file (and perhaps the comment adapted so it's obvious that it's helping use of flexbridge).

> Reviewed 1 of 1 files at r1, 1 of 2 files at r2.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/311#-:-MYjmqLD6s2AngwVib7L:b-m6p4tk)* status: :shipit: complete! all files reviewed, all discussions resolved

papeh:
> Firefox-Linux64 directory of .so files
FB was using these directly from the packages directory, since FB doesn't ship them. There are no .so files in output/DebugMono.

> HGRCPATH
I put that straight in the `flexbridge` script, if that's acceptable.

PR #324, opened 2021-06-17
Fix packaging

ermshiperete:
Allow to build package without running gitversion during binary package build. Instead run gitversion before creating the source
package which creates a json file. Then during binary package build we use the json file to populate the necessary properties.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/324)


marksvc:
> Allow to build package without running gitversion during binary package build. Instead run gitversion before creating the source
package which creates a json file. Then during binary package build we use the json file to populate the necessary properties.

This sounds great, thank you!

PR #328, opened 2021-08-10
Update vscode launch config

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/328)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -5,16 +5,16 @@
   "version": "0.2.0",
   "configurations": [
     {
-      "name": "Launch",
+      "name": "Launch FLExBridge",
       "type": "mono",
       "request": "launch",
-      "program": "${workspaceRoot}/output/DebugMono/FLExBridge.exe",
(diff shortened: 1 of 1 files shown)


.vscode/launch.json
```diff
@@ -29,9 +29,13 @@
         "-pipeID", "SendReceive${env:HOME}/fwrepo/fw/DistFiles/Projectsobtain"
       ],
       "env": {
+        // Send and receive to the QA LanguageForge server instead of production, if desired.
+        "LANGUAGEFORGESERVER": "-qa.languageforge.org",
```
ermshiperete, line 33: Is the dash in front of qa intentional?

ermshiperete (approved):
:lgtm:, one question though...

.vscode/launch.json
marksvc: Thank you. Yes. I think it may be combined with another string. It was copied this way from FieldWorks.

PR #329, opened 2021-08-10
doc: Add more Linux setup

marksvc:
- These items are needed for FW to launch FB. But really they should
probably be handled by another part of the build process rather than
done manually.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/329)

### DIFF
--- a/ReadMe.md
+++ b/ReadMe.md
@@ -47,6 +47,12 @@ cd build
 msbuild /t:Test FLExBridge.proj
 ```
 
+Additional setup for Linux:
+```bash
+cp -a flexbridge output/Debug/net461/
+cp -a DistFiles output/
+```
+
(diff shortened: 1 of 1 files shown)


ermshiperete (approved):

ermshiperete:
:lgtm:, although as you say that should be done automatically. I'd say it should go into `FLExBridge.csproj`

marksvc:
@ermshiperete  I had a go at this today. I started with FLExBridge.csproj but then saw that FLExBridge.proj had an existing spot that they fit well into. Needing this output/DistFiles/localizations seems like it may be a mistake somewhere, since it's redundant with output/Debug/net461/localizations .
I will send another PR.

PR #331, opened 2021-08-11
build: Copy files for Linux during build

marksvc:
(There is still an _error_ about localizations (missing), but at least FW can launch FB.)

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/331)

### DIFF
--- a/build/FLExBridge.proj
+++ b/build/FLExBridge.proj
@@ -108,11 +108,13 @@
 		<ChorusHubFiles Include="$(RootDir)/lib/$(Configuration)/ChorusHub.*"/>
 		<ConfigFiles Include="$(RootDir)/lib/$(Configuration)/*.dll.config"/>
 		<GeckoBrowserFiles Include="$(RootDir)/lib/$(Configuration)/SIL.Windows.Forms.GeckoBrowserAdapter.dll*"/>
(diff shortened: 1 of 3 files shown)


build/FLExBridge.proj
```diff
@@ -98,6 +98,8 @@
 		<MakeDir Directories="$(RootDir)/output/$(Configuration)/"/>
 		<MakeDir Directories="$(RootDir)/output/$(Configuration)/net461/"/>
 		<MakeDir Directories="$(RootDir)/output/$(Configuration)/net461/localizations/"/>
+		<MakeDir Directories="$(RootDir)/output/DistFiles"/>
```
ermshiperete: I wonder - do we really need these directories here, or is it looking for them in this place because the binaries are now one level lower in `net461`? In the latter case it would probably be better to adjust the code to look in the new place.

ermshiperete (approved):
:lgtm: - one question though.

build/FLExBridge.proj
marksvc: Right! Good thinking. I had a look for what's going on, but then found that at this point it is no longer necessary to copy DistFiles into output/. So that must have been fixed by earlier adjustments. 

I have changed the PR.

ermshiperete (approved):

PR #333, opened 2021-08-12
Fix LT-20773 crash using settings

jasonleenaylor:
* Use Nuget to bring back the GeckoBrowserAdapter that once
  lived in the lib folder

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/333)


papeh (approved):
If this breaks, CI builds should fail.

PR #339, opened 2021-09-10
launch: Optionally listen for debugger

marksvc:
- Usage: Set CHORUSDEBUGGING=true before launching.
- FB_MONO_ARGS and CHORUS_MONO_ARGS will allow other custom
commandline arguments to be externally set if helpful.
- Use non-default port 55502 for FlexBridge, and 55503 for
ChorusMerge, to not collide with others.
- Note that CHORUSDEBUGGING is also used by Chorus to print hg
debugging output. Setting in launch.json to set it for Chorus when
launching from launch.json, which should result both in hg debugging
output as well as listening for a debugger when ChorusMerge is later
run.

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/339)

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -36,6 +36,8 @@
         // Note that FLExBridge is run in the environment inherited from FieldWorks, such as set by FW environ.
         // Path to enviromnent settings for FieldWorks. This is handled by a mono wrapper script in /opt.
         "MONO_ENVIRON": "${workspaceRoot}/../fw/environ",
+        // Make Chorus print hg debugging info.
+        "CHORUSDEBUGGING": "true",
       },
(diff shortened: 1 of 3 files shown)


jasonleenaylor (commented):
I think I'd like to enable the attaching for FlexBridge and ChorusMerge via the same environment variable. Something that screams Send/Receive debugging. I also feel like I already created some such environment variable historically. 

> *[Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/339#-:-MjV58b2-6-dOZPcBJlT:b8x9q29)* status: 0 of 2 files reviewed, all discussions resolved

megahirt:
shall we close this old PR @marksvc ?

PR #341, opened 2021-09-23
Make build more robust

papeh:
- Depend on Chorus instead of Palaso in all places, so we always expect the same version
- Always use Platform=Any CPU, because it is the only valid platform
- Don't fail debug builds for missing localizations; the setup requires knowing a secret
- Replace hard-coded paths with variables for paths used multiple times in FLExBridge.proj

---
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/341)


jasonleenaylor (approved):
> Reviewed 2 of 2 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/341)* status: :shipit: complete! all files reviewed, all discussions resolved (waiting on @papeh)

PR #357, opened 2021-11-22
Add Flatpak FieldWorks GHA

marksvc:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/357)

### DIFF
--- a/.github/workflows/flatpak-fieldworks.yml
+++ b/.github/workflows/flatpak-fieldworks.yml
@@ -93,7 +93,7 @@ jobs:
           # Find the source stanza for 'flexbridge.git' and change its 'commit:' field to the commit id this build is checking.
           echo gs: ${GITHUB_SHA}
           perl -0777 -pi -e \
-            's/(type: git[\s\S]*?flexbridge\.git[\s\S]*?commit: )(.*)(\n[\s\S]*?- type:)/$1112233$3/gm' \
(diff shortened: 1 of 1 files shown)


ermshiperete:
I find it surprising to have a GHA that builds FieldWorks in the flexbridge repo. Wouldn't it be better to do it in the FieldWorks repo? I'd think it should be possible to do even though the FW changes go through Gerrit since we mirror the repo to GH.

marksvc:
The Flatpak package for FieldWorks depends on a number of different included components, one of which is Flexbridge. It will be useful to know if a change to Flexbridge is going to break or continue working in the FieldWorks Flatpak package-build. (Tho we might not choose for that to be a _blocking_ failure on a Flexbridge PR.)

This GHA is intended to build the FieldWorks Flatpak package, but with the proposed Flexbridge changes, showing what will happen if we later instruct the Flatpak package to use these new Flexbridge commits. 

Yes it is probably an odd GHA to see :-). But that's the thinking. What do you think about that? Maybe it's not clear enough that the point of the GHA is to check the impact of the Flexbridge changes on the FW Flatpak package-build? 

I do have a pre-merge build for proposed FieldWorks changes, that operates similarly to this GHA but that checks the proposed FieldWorks commit.

megahirt:
This old PR should likely be closed.  Would you agree @marksvc and @jasonleenaylor ?

PR #385, opened 2023-07-13
augotenerate binding redirects

papeh:
This may or may not fix anything immediately, but it is a good idea going forward.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/385)


ermshiperete (approved):
LGTM, although "augotenerate" is an interesting word :smile:

papeh:
Haha, good catch!

PR #387, opened 2023-07-18
Autogenerate ProductIdGuid for each build

papeh:
Since we are not patching, we don't need a predictable GUID.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/387)


papeh:
I tried this locally, and FLEx Bridge installed just fine over another version. Unfortunately, my locally build installer resulted in a FLEx Bridge that didn't run whether or not I specified a GUID, so I couldn't test end-to-end.

papeh:
Tested with TC builds; this works.

PR #398, opened 2024-04-18
Expose PutHumptyTogetherAgain method in LfMergeBridge

### LINKED ISSUE #782: Expose .fwdata file splitting/reassembling APIs in LfMergeBridge
This is a tracking issue to track work already done in https://github.com/sillsdev/flexbridge/pull/398. This is a prerequisite for https://github.com/sillsdev/languageforge-lexbox/issues/631 to be completed, but needs to be tracked separately from #631 so I can keep track of what work was finished in April vs. what was postponed to May.

rmunn:
The LexBox project wants to be able to [generate .fwdata files on the fly](https://github.com/sillsdev/languageforge-lexbox/issues/631), which will enable various features such as being able to view recent changes to a project in a nicer form than Mercurial diffs. In order to make this work, we need to expose the PutHumptyTogetherAgain method as a public method.

This PR seems the simplest way to do it, enabling access to that API from LfMergeBridge without changing the access granted to any other projects. That way we minimize the risk to FLEx and WeSay, as the code they use is not touched at all.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/398)

### DIFF
--- a/src/LfMergeBridge/LfMergeBridge.cs
+++ b/src/LfMergeBridge/LfMergeBridge.cs
@@ -71,5 +71,10 @@ namespace LfMergeBridge
 
 			return true;
 		}
+
+		public static void PutHumptyTogetherAgain(IProgress progress, bool writeVerbose, string mainFilePathname)
+		{
+			LibFLExBridgeChorusPlugin.DomainServices.FLExProjectUnifier.PutHumptyTogetherAgain(progress, writeVerbose, mainFilePathname);
+		}
 	}
 }
--- a/src/LfMergeBridge/LfMergeBridge.cs
+++ b/src/LfMergeBridge/LfMergeBridge.cs
@@ -72,7 +72,12 @@ namespace LfMergeBridge
 			return true;
 		}
 
-		public static void PutHumptyTogetherAgain(IProgress progress, bool writeVerbose, string mainFilePathname)
+		public static void DisassembleFwdataFile(IProgress progress, bool writeVerbose, string mainFilePathname)
+		{
+			LibFLExBridgeChorusPlugin.DomainServices.FLExProjectSplitter.PushHumptyOffTheWall(progress, writeVerbose, mainFilePathname);
+		}
+
+		public static void ReassembleFwdataFile(IProgress progress, bool writeVerbose, string mainFilePathname)
 		{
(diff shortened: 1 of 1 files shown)


jasonleenaylor (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/398)* status: 0 of 1 files reviewed, all discussions resolved

___
*[`src/LfMergeBridge/LfMergeBridge.cs` line 75 at r1](https://reviewable.io/reviews/sillsdev/flexbridge/398#-NvmG8n-DXQkciwC4CAm:-NvmG8n-DXQkciwC4CAn:b-tkaakj) ([raw file](https://github.com/sillsdev/flexbridge/blob/77899f1d73cc58306bb06037233bea387d9b1ca6/src/LfMergeBridge/LfMergeBridge.cs#L75)):*
> ```Smalltalk
> 		}
> 
> 		public static void PutHumptyTogetherAgain(IProgress progress, bool writeVerbose, string mainFilePathname)
> ```

This is an opportunity, should you want one, to adjust the name.



_Code quote:_
```Smalltalk
PutHumptyTogetherAgain
```

rmunn (commented):
> *[Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/398)* status: 0 of 1 files reviewed, all discussions resolved (waiting on @jasonleenaylor)

___
*[`src/LfMergeBridge/LfMergeBridge.cs` line 75 at r1](https://reviewable.io/reviews/sillsdev/flexbridge/398#-NvmG8n-DXQkciwC4CAm:-NwD6149AnTslQp9v-DX:b5xfba0) ([raw file](https://github.com/sillsdev/flexbridge/blob/77899f1d73cc58306bb06037233bea387d9b1ca6/src/LfMergeBridge/LfMergeBridge.cs#L75)):*

This is an opportunity, should you want one, to adjust the name.
</blockquote></details>

Good point. I've pushed a new commit to use more boring, but easier to understand, names. And while I'm at it, I've decided to expose the reverse operation too; I have no need for it *yet* in the code I'm about to write in LexBox, but could definitely see a need for this in future features.

rmunn:
I'm assuming Jason would have mentioned any objections if he had them; since this affects nothing except the code that interfaces with LfMerge, I'm going to go ahead and merge it now.

PR #397, opened 2024-04-17
Signing CD Action

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/397)

PR #399, opened 2024-04-26
Fix Chorus.Mercurial package issue on Linux

rmunn:
Version 6.5.1.25 of the SIL.Chorus.Mercurial package had a few missing files that cause test failures in some circumstances; 6.5.1.29 fixes that omission.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/399)

### DIFF
--- a/.github/workflows/ci-cd.yml
+++ b/.github/workflows/ci-cd.yml
@@ -5,16 +5,23 @@ on:
   pull_request:
     branches: ["develop", "master"]
   workflow_dispatch:
+    inputs:
+      releaseType:
+        description: 'Alpha, Beta, Stable used to stamp release artifacts'
+        required: true
+        default: 'Alpha'
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
 
 jobs:
-  windows_debug_build_and_test:
+  build_and_test:
     env:
       CROWDIN_API_KEY: ${{ secrets.CROWDIN_API_KEY }}
-    name: Build Debug and run Tests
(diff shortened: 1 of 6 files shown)


rmunn:
I should note that the missing files are compiled C extensions, which I compiled for Python 3.8 and above since those are the currently-supported versions of Python (3.7 went out of support in October 2023, earlier versions were of course earlier than that). But our TeamCity build agent for linux64, which has been failing to build ever since the SIL.Chorus.Mercurial package was updated to 6.5.1 (which requires Python 3), is running Ubuntu Xenial, where the Python version was 3.5.

If we want to get this building on TeamCity, we might need to update the build runner over there to Ubuntu 22.04. (Or even 24.04, which just came out today). Either that, or move the NuGet packaging-and-release build to GitHub Actions.

(I don't care where it builds, I just want it to be able to consume a NuGet package with the changes from https://github.com/sillsdev/flexbridge/pull/398).

rmunn:
PR 397 [changes the package version to `6.*`](https://github.com/sillsdev/flexbridge/pull/397/files#diff-d071cf6f93578667b04cab0f957540f914291c0a7f6f0795bcb960948bb3dd96R15) so that it will automatically pull in any future Mercurial packaging fixes. I'll change this PR to do the same so that PR 397 won't have any merge conflicts to deal with once it's ready to merge.

rmunn:
@jasonleenaylor - Any objections to my merging this before PR #397 is fully ironed out? I've been holding off so as to not cause merge conflicts, but perhaps that's a little silly for a one-line change.

jasonleenaylor:
Robin, sorry I didn't see your comment here earlier. After merging this with the results of the ci-cd build there are no changes left in this commit, so I'm going to close the PR.

PR #401, opened 2024-06-24
Faster ChorusNotes handling for LfMerge

rmunn:
Goes along with https://github.com/sillsdev/LfMerge/pull/338, which will be mergeable once this PR is merged and a new NuGet package is created with the LfMergeBridge code changes.

This PR moves several POCO classes from LfMerge into LfMergeBridge. This allows these classes to be passed by reference into the ChorusNotes action handler, saving two large JSON serialization steps, as suggested in https://github.com/sillsdev/LfMerge/issues/329.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/401)

### DIFF
--- a/src/LfMergeBridge/LanguageForgeGetChorusNotesActionHandler.cs
+++ b/src/LfMergeBridge/LanguageForgeGetChorusNotesActionHandler.cs
@@ -45,11 +45,10 @@ namespace LfMergeBridge
 			ProjectDir = Path.GetDirectoryName(pOption);
 
 			List<LfComment> commentsFromLF;
-			if (LfMergeBridge.ExtraData.TryGetValue(options, out var extraData) && extraData is GetChorusNotesInput inputData)
+			if (LfMergeBridge.ExtraInputData.TryGetValue(options, out var extraData) && extraData is GetChorusNotesInput inputData)
 			{
 				Console.WriteLine("Extra data passed in, of the right type");
 				commentsFromLF = inputData.LfComments;
-				LfMergeBridge.ExtraData.Remove(options);
 				hadExtraData = true;
 			}
 			else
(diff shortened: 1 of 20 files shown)


rmunn:
About to rebase on top of `develop` now that the CI signing PR has been merged.

src/LfMergeBridge/LanguageForgeGetChorusNotesActionHandler.cs
```diff
@@ -37,90 +38,125 @@ internal sealed class LanguageForgeGetChorusNotesActionHandler : IBridgeActionTy
...
+				serializedReplies.Append(JsonConvert.SerializeObject(lfReplies));
+				LfMergeBridgeUtilities.AppendLineToSomethingForClient(ref somethingForClient, serializedReplies.ToString());
 
-			var serializedStatusChanges = new StringBuilder("New status changes on comments already in LF: ");
-			serializedStatusChanges.Append(JsonConvert.SerializeObject(lfStatusChanges));
-			LfMergeBridgeUtilities.AppendLineToSomethingForClient(ref somethingForClient, serializedStatusChanges.ToString());
+				var serializedStatusChanges = new StringBuilder("New status changes on comments already in LF: ");
+				serializedStatusChanges.Append(JsonConvert.SerializeObject(lfStatusChanges));
+				LfMergeBridgeUtilities.AppendLineToSomethingForClient(ref somethingForClient, serializedStatusChanges.ToString());
+			}
```
megahirt: I clarified this in person, and I don't think we want to keep the old implementation around.  We are switching to the new implementation that avoids string serialization of comments.  This is an opportunity to reduce complexity in our code.

megahirt (approved):
The worked to demonstrate that this new code works is complicated, so I won't verify this personally.  Robin has conveyed to me in person how this fits together and so I approve of this PR.  Looking forward to seeing it integrated in time!

src/LfMergeBridge/LanguageForgeGetChorusNotesActionHandler.cs
rmunn: Done in commit b572aa50.

rmunn:
I'm a bit nervous about having added GenerateResourceUsePreserializedResources to so many .csproj files; a cursory reading of https://www.github.com/dotnet/msbuild/issues/4704 suggests that it might cause the build to fail for net461 unless the System.Resources.Extensions package is also referenced by the .csproj file. It builds correctly without that package reference on my machine, though, and those reports are several years old, so hopefully that is not a problem any more. But if there are errors building for net461, I will open another PR to either add the System.Resources.Extensions NuGet package or else remove the GenerateResourceUsePreserializedResources property.

Noting this now because I'm about to merge this PR, and I want to make sure I can easily find the solution if it turns out merging the PR breaks the build of the `develop` branch.

PR #402, opened 2024-06-25
Don't serialize JSON in WriteToChorusNotes handler

rmunn:
Companion PR to #401. Once this is merged, https://github.com/sillsdev/LfMerge/pull/338 should follow.

Similarly to what we just did with the GetChorusNotes handler, we can skip serializing JSON in the WriteToChorusNotes handler as well.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/402)


megahirt (approved):
LGTM
Nice to see the old TODOs removed since we are using a more modern C# version!

PR #405, opened 2024-07-01
Add NuGet packaging step to CI/CD workflow

### LINKED ISSUE #400: Move FlexBridge NuGet packaging build to GitHub Actions
The build that creates the NuGet package(s) for FlexBridge is currently running on TeamCity. It's been failing to build since [the commit that updated SIL.Chorus.Mercurial to 6.5.1](https://github.com/sillsdev/flexbridge/commit/a41bce8a141666178a52ef973b4560bce44f7ad8). The reason, at least for the Linux64 build, is because the TeamCity runner is still on Ubuntu Xenial, which has Python 3.5. Mercurial, starting with release 6.2, [supports Python 3.6 (or above) only](https://wiki.mercurial-scm.org/Release6.2).

We could fix this by creating a new TeamCity Linux build agent, but we're trying to move away from TeamCity. The better solution would be to move the NuGet packaging build to GitHub Actions.

rmunn:
Fixes #400.

This PR moves NuGet publishing to GitHub actions and removes the Appveyor CI.  Appveyer was only responsible for NuGet publishing, so this consolidates CI to only use GHA.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/405)


rmunn:
Note to self: Might be a good idea to push PR builds to https://int.nugettest.org/ (API endpoint is at https://apiint.nugettest.org/v3/index.json rather than the standard one). Microsoft makes no guarantees about data preservation, but I expect they would stick around for at least a few hours before being deleted, which would allow other code to consume them for testing purposes. For example, I could add a new feature to LfMergeBridge in a PR to the flexbridge repo, then consume a NuGet package in LfMerge and be able to test the code right away, without having to mess with local NuGet feeds.

See https://learn.microsoft.com/en-us/nuget/nuget-org/nuget-org-faq#managing-packages-on-nuget-org for where I got the idea from.

rmunn:
Workflow failed because our API key isn't registered on int.nugettest.org. We may want to get that taken care of at some point, because it would be really nice to have the ability to reference NuGet packages built from PRs.

megahirt (approved):
LGTM

PR #413, opened 2024-11-27
Add a test result publishing step

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/413)

PR #416, opened 2024-11-29
Update LfMergeBridge tests to new comments API

### LINKED ISSUE #415: Failing LfMergeBridge tests
When https://github.com/sillsdev/flexbridge/pull/401 was merged, changing the API for calling LfMergeBridge, I forgot to update the tests to match the change in API.

rmunn:
New API (introduced in #401 and #402) for LfMergeBridge passes LfComment lists directly rather than serialized to JSON, so the tests need to do the same.

Fixes #415.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/416)


rmunn:
CI is passing despite test failures due to wrong versions of SIL.Core, etc., and yet the CI is marking them as passing. I believe #413 is supposed to fix that; I'll change this PR to merge into #413.

hahn-kev (approved):

PR #420, opened 2024-12-03
Don't try to run tests from SIL.LCModel.Utils.Tests

jasonleenaylor:
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/420)

### DIFF
--- a/build/FLExBridge.proj
+++ b/build/FLExBridge.proj
@@ -1,7 +1,6 @@
 <Project xmlns="http://schemas.microsoft.com/developer/msbuild/2003" DefaultTargets="Build;Pack">
 	<PropertyGroup>
-		<RootDir Condition="'$(teamcity_build_checkoutDir)' == '' And '$(RootDir)'==''">$(MSBuildProjectDirectory)/..</RootDir>
-		<RootDir Condition="'$(teamcity_build_checkoutDir)' != ''">$(teamcity_build_checkoutDir)</RootDir>
+		<RootDir>$(MSBuildProjectDirectory)/..</RootDir>
 		<Configuration Condition="'$(Configuration)'==''">Debug</Configuration>
 		<Platform>Any CPU</Platform>
 		<Solution>FLExBridge.sln</Solution>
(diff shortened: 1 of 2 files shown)


build/FLExBridge.proj
```diff
@@ -203,19 +194,14 @@
...
-			TestInNewThread="false"
-			ExcludeCategory="$(ExtraExcludeCategories)"
-			WorkingDirectory="$(RootDir)/output/$(Configuration)/$(TargetFramework)"
-			Process="single"
-			Verbose="true"
-			UseNUnit3Xml="false"
-			OutputXmlFile="$(RootDir)/output/$(Configuration)/$(TargetFramework)/TestResults.xml"
-			TeamCity="$(TeamCity)"/>
+		<!-- Loop over each TestAssembly and execute dotnet test -->
+		<Exec Command="dotnet test --no-build --verbosity detailed --logger &quot;trx;LogFileName=%(TestAssemblies.Filename)TestResults.trx&quot; --filter &quot;TestCategory!=$(ExtraExcludeCategories)&quot; &quot;%(TestAssemblies.FullPath)&quot;"
```
hahn-kev, line 203: escaping stuff sucks, could you do this instead?
```suggestion
		<PropertyGroup>
			<_RunTestsCommand>dotnet test --no-build --verbosity detailed --logger "trx;LogFileName=%(TestAssemblies.Filename)TestResults.trx" --filter "TestCategory!=$(ExtraExcludeCategories)" "%(TestAssemblies.FullPath)"</_RunTestsCommand>
		</PropertyGroup>
		<!-- Loop over each TestAssembly and execute dotnet test -->
		<Exec Command="$(_RunTestsCommand)"
```
I'm not too familiar with msbuild so I'm not sure if that would work but it would be much more maintinable

hahn-kev (approved):
looks good to me, I left a suggestion, but I'm not super sure if it'll work so feel free to discard it. I tried it and got an error I didn't really understand, but I'm not sure if that's just my environment.

build/FLExBridge.proj
jasonleenaylor: Yeah, escaping stuff isn't my favorite, but in this case I think it is necessary.
Your suggestion wouldn't work because it would defeat the batching. Actually, the error you got is probably because the metadata expansion isn't valid in a PropertyGroup.
https://learn.microsoft.com/en-us/visualstudio/msbuild/item-metadata-in-task-batching?view=vs-2022

PR #429, opened 2026-07-01
Fix LT-19237: preserve Data Notebook fields merging record-type layouts

jasonleenaylor:
The Data Notebook writes one layout per record type, all sharing the same class/type/name and differing only by choiceGuid. Leaving choiceGuid out of the merge key let Send/Receive confuse one record type's layout for another's, dropping most fields from a record type.

- - -
This change is [Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/429)

### DIFF
--- a/src/LibFLExBridge-ChorusPlugin/Handling/ConfigLayout/CustomLayoutDataCollectorMethod.cs
+++ b/src/LibFLExBridge-ChorusPlugin/Handling/ConfigLayout/CustomLayoutDataCollectorMethod.cs
@@ -15,9 +15,15 @@ namespace LibFLExBridgeChorusPlugin.Handling.ConfigLayout
 		internal static Dictionary<string, byte[]> GetDataFromRevision(FileInRevision revision, HgRepository repository)
 		{
 			var doc = XDocument.Parse(revision.GetFileContents(repository));
+			// LT-19237: include choiceGuid in the key. The Data Notebook emits one <layout> per
+			// record type, all sharing class/type/name and differing only by choiceGuid; without it
+			// the keys collide and ToDictionary throws a duplicate-key ArgumentException. choiceGuid
+			// is optional, so layouts without it key on class+type+name+"" (byte-identical to the old
+			// key). Matches the merge partner key in CustomLayoutMergeStrategiesMethod.
 			var data = doc.Root.Elements("layout")
 				.ToDictionary(layoutElement =>
-							  layoutElement.Attribute("class").Value + layoutElement.Attribute("type").Value + layoutElement.Attribute("name").Value,
+							  layoutElement.Attribute("class").Value + layoutElement.Attribute("type").Value + layoutElement.Attribute("name").Value
+							  + (layoutElement.Attribute("choiceGuid")?.Value ?? ""),
 					layoutElement => LibTriboroughBridgeSharedConstants.Utf8.GetBytes(layoutElement.ToString()));
 
 			var layoutTypeElement = doc.Root.Element("layoutType");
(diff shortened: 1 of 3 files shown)


imnasnainaec:
[Devin](https://app.devin.ai/review/sillsdev/flexbridge/pull/429) bug, flag, and info:

src/LibFLExBridge-ChorusPlugin/Handling/ConfigLayout/CustomLayoutDataCollectorMethod.cs:18
**Two-way diff crashes with duplicate key when multiple layouts share the same class, type, and name**
The dictionary key used to collect layout data for diffs is built from only class+type+name (`CustomLayoutDataCollectorMethod.cs:20`), so when two layouts share those attributes but differ by choiceGuid, `ToDictionary` throws a duplicate-key exception.

Impact: Any two-way diff on a file containing Data Notebook record-type layouts (the exact scenario this PR fixes for merges) will crash instead of showing changes.

<details>
<summary>The data-collector key was not updated to match the new merge key</summary>

The PR correctly adds `choiceGuid` to the merge partner finder key at `CustomLayoutMergeStrategiesMethod.cs:37`, so 3-way merges now distinguish layouts that share class/type/name but have different choiceGuid values.

However, `CustomLayoutDataCollectorMethod.cs:18-21` still builds its dictionary key as `class + type + name` without including `choiceGuid`. When the Data Notebook emits multiple `<layout>` elements like:

```xml
<layout class='RnGenericRec' type='detail' name='Normal' choiceGuid='guid-A' .../>
<layout class='RnGenericRec' type='detail' name='Normal' choiceGuid='guid-B' .../>
```

the `ToDictionary` call produces `"RnGenericRecdetailNormal"` for both elements and throws `System.ArgumentException` ("An item with the same key has already been added").

The fix should append `choiceGuid` (when present) to the dictionary key, e.g.:
```
layoutElement.Attribute("class").Value + layoutElement.Attribute("type").Value + layoutElement.Attribute("name").Value + (layoutElement.Attribute("choiceGuid")?.Value ?? "")
```
</details>

---

src/LibFLExBridge-ChorusPlugin/Handling/ConfigLayout/CustomLayoutMergeStrategiesMethod.cs:37
**Merge key change is a behavioral contract change for all layout files, not just Data Notebook**
The merge strategy for `<layout>` elements is shared across all `.fwlayout` files, not just Data Notebook layouts. Adding `choiceGuid` to the key at `CustomLayoutMergeStrategiesMethod.cs:37` changes the matching behavior for every layout file processed by `FieldWorksConfigurationLayoutTypeHandlerStrategy`. While the PR comment at lines 32-36 correctly notes that layouts without a `choiceGuid` attribute will all key on a null value and "continue to match each other exactly as before," this depends on how `FindByMultipleKeyAttributes` in the Chorus library treats missing attributes. If it returns null/empty for a missing attribute, layouts without `choiceGuid` will still match each other correctly. This assumption appears sound based on the existing pattern at `FieldWorksMergeServices.cs:77` where `FindByMultipleKeyAttributes` is used with attributes that may not always be present, but since Chorus is an external NuGet dependency (`6.0.0-beta0063`), the exact behavior couldn't be source-verified locally.

---

src/LibFLExBridge-ChorusPluginTests/Handling/ConfigLayout/FieldWorksCustomLayoutTypeHandlerTests.cs:327
**Test covers the specific reproduction scenario well but does not test the diff path**
The new test `Merge_RecordTypeLayoutsAreMatchedByChoiceGuid_LT19237` at line 327 thoroughly verifies the 3-way merge scenario with reversed ordering and independent edits to different record-type layouts. However, there is no corresponding test for the 2-way diff path (`Find2WayDifferences`), which exercises `CustomLayoutDataCollectorMethod.GetDataFromRevision` — the method that still uses only class+type+name as its dictionary key. A test calling `Find2WayDifferences` on a file with multiple layouts sharing class/type/name but differing by choiceGuid would immediately expose the crash reported as BUG-0001.

mark-sil (approved):
> @mark-sil reviewed 3 files and all commit messages.
> *[Reviewable](https://reviewable.io/reviews/sillsdev/flexbridge/429)* status: <span>:shipit:</span> complete! all files reviewed, all discussions resolved (waiting on jasonleenaylor).

